PR #165, opened 2023-04-05
log column encoding in materialize

matzemathics:


mmarx (approved):
Looks good, but let's see what the discussion on log levels leads to ;)

PR #187, opened 2023-04-19
Add TrieScanPrune operator

rlwww:


src/physical/columnar/operations/columnscan_prune.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// Index this column scan corresponds to in the input trie
+    column_scan_index: usize,
+    reference_scan: &'a ColumnScanCell<'a, T>,
+}
+
+impl<'a, T> ColumnScanPrune<'a, T>
+where
+    T: 'a + ColumnDataType,
+{
+    /// Constructs a new column scan. This should not be called directory, but only by `TrieScanPrune`
```
mmarx, line 26: `directory` -> “directly”?

src/physical/columnar/operations/columnscan_prune.rs
```diff
@@ -0,0 +1,87 @@
...
+        loop {
+            match self.next() {
+                Some(next_value) => {
+                    if next_value >= minimum_value {
+                        return Some(next_value);
+                    }
+                }
+                None => return None,
+            }
+        }
```
mmarx, line 66: isn't this just `self.find(|next_value| next_value >= minimum_value)`?

mmarx (commented):
I have two questions, but the rest looks good!

src/physical/columnar/operations/columnscan_prune.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+impl<'a, T> Iterator for ColumnScanPrune<'a, T>
+where
+    T: 'a + ColumnDataType,
+{
+    type Item = T;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        unsafe { (*self.state.get()).get_next_value(self.column_scan_index) };
```
monsterkrampe, line 47: Since you have `&mut self` couldn't you do `self.state.get_mut().get_next_value(...)` without the `unsafe` block? Maybe I'm missing something.

src/physical/tabular/operations/triescan_prune.rs
```diff
@@ -0,0 +1,473 @@
...
+    #[inline]
+    pub fn is_column_peeked(&self, index: usize) -> bool {
+        self.column_peeks[index]
+    }
+
+    #[inline]
+    unsafe fn get_input_trie_current_value(&mut self, index: usize) -> Option<StorageValueT> {
+        let scan = self.input_trie_scan.get_scan(index).unwrap();
+
+        unsafe { &mut *scan.get() }.current()
```
monsterkrampe, line 119: Again, since you get `&mut self`, couldn't you use `get_mut` on the `UnsafeCell` without the `unsafe` block?
(Same comment for the method below.)

src/physical/tabular/operations/triescan_prune.rs
```diff
@@ -0,0 +1,473 @@
...
+            output_column_scans,
+            target_types,
+        }
+    }
+}
+
+impl<'a> TrieScan<'a> for TrieScanPrune<'a> {
+    fn up(&mut self) {
+        unsafe {
+            (*self.state.get()).external_up();
```
monsterkrampe, line 281: Again, `get_mut` might help getting rid of the `unsafe` block. (Same in methods below.)

monsterkrampe (changes requested):
Not sure if this really works but I think in some places one could replace `unsafe` blocks by calls to `get_mut` on `UnsafeCell`s since one already has a mutable reference to them anyway.

mmarx:
It's also already fixed on `main`, so just rebasing should get rid of it.

mmarx (changes requested):
Found a typo in the docs, looks good otherwise!

mmarx (approved):

PR #188, opened 2023-04-20
Actions/build main

ellmau:


.github/workflows/release.yml
```diff
@@ -0,0 +1,28 @@
...
+          - target: x86_64-apple-darwin
+            archive: zip
+    steps:
+      - uses: actions/checkout@master
+      - name: Compile and release
+        uses: rust-build/rust-build.action@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # needs to be provided to change assets
+          RUSTTARGET: ${{ matrix.target }}
+          EXTRA_FILES: "README.md LICENSE"
```
mmarx, line 27: Still has the wrong files in it.

.github/workflows/release.yml
```diff
@@ -0,0 +1,28 @@
...
+      matrix:
+        include:
+          - target: x86_64-pc-windows-gnu
+            archive: zip
+          - target: x86_64-unknown-linux-musl
+            archive: tar.gz
+          - target: x86_64-apple-darwin
+            archive: zip
+    steps:
+      - uses: actions/checkout@master
```
mmarx, line 21: Please use a stable release, i.e., `actions/checkout@v3`

.github/workflows/release.yml
```diff
@@ -0,0 +1,28 @@
...
+      - name: Compile and release
+        uses: rust-build/rust-build.action@master
```
mmarx, line 23: Again, don't track `master` here. Should be at least `@v1`.

mmarx (changes requested):

.github/workflows/release.yml
```diff
@@ -18,11 +18,11 @@ jobs:
           - target: x86_64-apple-darwin
             archive: zip
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@v3
       - name: Compile and release
-        uses: rust-build/rust-build.action@master
```
mmarx: Patch level is definitely too fine-grained, we don't want to update the action on _every_ release of `rust-build.action`. `@v1` should be fine.

mmarx (changes requested):

PR #191, opened 2023-04-20
Performance/joins

aannleax:
Improves the performance of existential rule reasoning.

Mainly does so by:

- Fixing a bug in the variable order computation (where variables have been sorted alphabetically by accident)
- Using a separate variable order for the head join as for the body join

Additionally includes the functionality to not fully materialize unnecessary layers of `TrieScan`, though it did not have any noticeable impact on the performance.


src/logical/program_analysis/analysis.rs
```diff
@@ -91,9 +98,76 @@ fn get_fresh_rule_predicate(rule_index: usize) -> Identifier {
...
+    );
+
+    let temp_program = Program::new(
+        None,
+        HashMap::new(),
+        Vec::new(),
+        vec![temp_rule.clone()],
+        Vec::new(),
+        HashMap::new(),
+    );
```
mmarx, line 155:
```rust
let temp_program = vec![temp_rule.clone()].into();
```

mmarx (changes requested):
Just some minor cleanup, looks good otherwise!

PR #201, opened 2023-04-23
Check constants for type consistency

### LINKED ISSUE #196: Type inference for constants in facts and rules
Using constants in rules or facts does currently not lead to any type restrictions, i.e. the following will not lead to a user error:
```
@declare less_than(UnsignedInteger, UnsignedInteger) .

less_than(1, 2) .
substring(a, ab) .

result(?X, ?Y) :- less_than(?X, ?Y), substring(?X, ?Y) .
```

Instead we get
```
thread 'main' panicked at 'Only NumericLiteral::Integer terms are supported for logical type UnsignedInteger!', src/logical/types.rs:105:26
```

aannleax:
Checks whether a `Term` appearing in fact or rule can be converted into the needed logical type, and outputs a user error if this is not the case.

Also fixes a bug where the logical types would not be correctly initialized with the user provided types.

Closes #196.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,7 +1,8 @@
 //! Error-handling module for the crate
 
 use crate::{
-    io::parser::ParseError, logical::types::LogicalTypeEnum,
+    io::parser::ParseError,
+    logical::{model::Term, types::LogicalTypeEnum},
     physical::datatypes::float_is_nan::FloatIsNaN,
 };
 use thiserror::Error;
@@ -43,6 +44,9 @@ pub enum Error {
     /// Conflicting type declarations
     #[error("Conflicting type declarations. Predicate \"{0}\" at position {1} has been inferred to have the conflicting types {2} and {3}.")]
     InvalidRuleConflictingTypes(String, usize, LogicalTypeEnum, LogicalTypeEnum),
+    /// Conflicting type declarations
+    #[error("Conflicting type declarations. The term \"{0}\" cannot be converted to a {1}.")]
+    InvalidRuleTermConversion(Term, LogicalTypeEnum),
     /// Unsupported feature: Negation
     #[error("Negation is currently unsupported.")]
     UnsupportedFeatureNegation,
--- a/src/logical/model.rs
+++ b/src/logical/model.rs
@@ -2,6 +2,7 @@
 
 use std::{
     collections::{HashMap, HashSet},
+    fmt::Debug,
     ops::Neg,
     path::{Path, PathBuf},
 };
@@ -34,6 +35,12 @@ impl Identifier {
     }
 }
 
+impl std::fmt::Display for Identifier {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        std::fmt::Display::fmt(&self.name(), f)
+    }
+}
+
 /// Variable occuring in a rule
 #[derive(Debug, Eq, PartialEq, Hash, Clone, PartialOrd, Ord)]
 pub enum Variable {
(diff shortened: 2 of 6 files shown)


src/error.rs
```diff
@@ -1,7 +1,8 @@
 //! Error-handling module for the crate
 
 use crate::{
-    io::parser::ParseError, logical::types::LogicalTypeEnum,
+    io::parser::ParseError,
+    logical::{model::Term, types::LogicalTypeEnum},
     physical::datatypes::float_is_nan::FloatIsNaN,
```
ellmau, line 6: Following the [style guide for imports](https://doc.rust-lang.org/1.0.0/style/style/imports.html), we should reorder the imports as follows:
```rust
use thiserror::Error;

use crate::{
    io::parser::ParseError,
    logical::{model::Term, types::LogicalTypeEnum},
    physical::datatypes::float_is_nan::FloatIsNaN,
};
```

src/logical/program_analysis/analysis.rs
```diff
@@ -467,6 +475,60 @@ impl Program {
         Ok(())
     }
 
+    // Check if there is a constant that cannot be converted into the type of
+    // the variable/predicate position it is compared to.
```
ellmau, line 479:
```suggestion
    /// Check if there is a constant that cannot be converted into the type of
    /// the variable/predicate position it is compared to.
```
Use doc-comments instead

src/logical/model.rs
```diff
@@ -34,6 +35,12 @@ impl Identifier {
     }
 }
 
+impl std::fmt::Display for Identifier {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        std::fmt::Display::fmt(&self.name(), f)
```
monsterkrampe, line 40: I think the common way of implementing this is:
```suggestion
        write!(f, "{}", self.name())
```
But maybe I'm wrong; I'm fine with both.

src/logical/model.rs
```diff
@@ -52,6 +59,12 @@ impl Variable {
     }
 }
 
+impl std::fmt::Display for Variable {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        std::fmt::Display::fmt(&self.name(), f)
```
monsterkrampe, line 64: Same as above.

src/logical/model.rs
```diff
@@ -65,6 +78,17 @@ pub enum Term {
     RdfLiteral(RdfLiteral),
 }
 
+impl std::fmt::Display for Term {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match &self {
+            Term::Constant(term) => std::fmt::Display::fmt(term, f),
+            Term::Variable(term) => std::fmt::Display::fmt(term, f),
+            Term::NumericLiteral(term) => std::fmt::Display::fmt(term, f),
+            Term::RdfLiteral(term) => std::fmt::Display::fmt(term, f),
```
monsterkrampe, line 87: Same as above.

ellmau (approved):
Two minor nitpicks for better code-formatting and documentation, LGTM otherwise.

src/logical/program_analysis/analysis.rs
```diff
@@ -377,12 +377,20 @@ impl Program {
...
+            .parsed_predicate_declarations()
+            .into_iter()
+            .map(|(predicate, types)| (predicate, types.into_iter().map(Some).collect()))
             .collect();
 
+        // Set all predicates that did not receive explicit type information to `None` which represents unknown.
+        for (predicate, arity) in all_predicates {
+            if let Entry::Vacant(entry) = predicate_types.entry(predicate.clone()) {
+                entry.insert(vec![None; *arity]);
+            }
```
monsterkrampe, line 391:
```suggestion
            predicate_types.entry(predicate.clone()).or_insert(vec![None; *arity]);
```

monsterkrampe (changes requested):
Looks good! I only have some minor comments.

aannleax:
Implemented your suggestions. Thanks :)

PR #203, opened 2023-04-24
Fixed bug in seminaive join of head and changed selection strategy

aannleax:
Improved the performance of existential rule reasoning by:
* Changing what are considered "old" and "new" tables in the seminaive join of the rule head
* Changing the rule selection strategy to Graph-based

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -12,7 +12,11 @@ use nemo::{
     },
     logical::{
         execution::{
-            selection_strategy::strategy_round_robin::StrategyRoundRobin, ExecutionEngine,
+            selection_strategy::{
+                dependency_graph::graph_positive::GraphConstructorPositive,
+                strategy_graph::StrategyDependencyGraph, strategy_round_robin::StrategyRoundRobin,
+            },
+            ExecutionEngine,
         },
         model::Program,
     },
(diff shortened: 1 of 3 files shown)


src/app.rs
```diff
@@ -162,7 +166,11 @@ impl CliApp {
 
         self.prevent_accidential_overwrite(&app_state)?;
 
-        let mut exec_engine = ExecutionEngine::<StrategyRoundRobin>::initialize(app_state.program)?;
+        // let mut exec_engine = ExecutionEngine::<StrategyRoundRobin>::initialize(app_state.program)?;
```
mmarx, line 169: This can be removed, right?

mmarx (approved):

src/logical/execution/planning/plan_head_restricted.rs
```diff
@@ -149,10 +149,16 @@ impl HeadStrategy for RestrictedChaseStrategy {
 
         // 1. Compute the table "New Satisfied Matches"
 
+        let step_last_applied = if rule_info.step_last_applied == 0 {
+            0
+        } else {
+            rule_info.step_last_applied + 1
```
monsterkrampe, line 155: Can you leave a comment on why it is necessary to add `1` here? At least to me this is not obvious :D

monsterkrampe: I thinks it's not super clear to me yet but having a comment should certainly help in the future (to at least raise awareness that this part is non-trivial).

monsterkrampe (approved):

mmarx (approved):

PR #204, opened 2023-04-25
Improve description of output CLI parameters

### LINKED ISSUE #175: Change command-line parameter name for output directory
Instead of the current -o/--output parameters, we should use the following:
```
-D, --output-dir=<DIR>   Specify directory for output files.
```
This should default to "." This will align that option with Souffle and Formulog. As part of the change to use . as default directory, the behavior of ```--overwrite-results``` should change to overwrite only the files that are newly generated, and to leave all other files in the directory unchanged. In general, the behaviour of deleting everything in a directory with this option seems quite dangerous and easy to overlook.

Later, there should also be a revision of the current ```-s``` option, but this depends on the availability of output directives as per #174.

monsterkrampe:
closes #175

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,12 +1,15 @@
 {
   "nodes": {
     "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
       "locked": {
-        "lastModified": 1678901627,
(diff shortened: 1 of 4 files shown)


ellmau (approved):
LGTM

PR #205, opened 2023-04-25
Normalize constants when parsing from DSV

### LINKED ISSUE #199: Unify handling of angled brackets in CSV and rules files
Currently, the parser interprets the expression `constant` in a rule or fact as an IRI and normalizes it to `<constant>`. However, the same does not happen when parsing CSV files. This has the annoying effect that constants used in input files cannot be referenced by rules unless they are enclosed by angled brackets.

monsterkrampe:
closes #199 
Bare names now support spaces; I'm not sure if we want that. Some CSV Test was assuming that spaces are allowed so I changed it.


src/io/builder_proxy.rs
```diff
@@ -33,10 +36,20 @@ impl ColumnBuilderProxy for StringColumnBuilderProxy {
...
+        let parsed = parse_dsv_constant(string)
+            .map(|(remainder, parsed)| {
+                debug_assert!(remainder.is_empty()); // the whole value should be parsed
+                parsed
+            })
+            .unwrap_or(string.to_string());
+
         self.value = Some(
             dictionary
                 .expect("ProxyStringColumnBuilder expects a Dictionary to be provided!")
```
mmarx, line 51: `ProxyStringColumnBuilder` ~> `StringColumnBuilderProxy` ;)

src/io/builder_proxy.rs
```diff
@@ -33,10 +36,20 @@ impl ColumnBuilderProxy for StringColumnBuilderProxy {
...
     fn add(&mut self, string: &str, dictionary: Option<&mut Dict>) -> Result<(), Error> {
         self.commit();
+
+        // TODO: parsing of DSV and Program should be unified
+        // TODO: DSV parsing should depend on logical types
+        let parsed = parse_dsv_constant(string)
+            .map(|(remainder, parsed)| {
+                debug_assert!(remainder.is_empty()); // the whole value should be parsed
+                parsed
+            })
```
mmarx, line 46: Could also use `all_input_consumed` here?

src/io/parser.rs
```diff
@@ -75,6 +75,34 @@ pub fn multispace_or_comment1(input: &str) -> IntermediateResult<()> {
...
+            alpha1,
+            opt(pair(
+                many0(tag(" ")),
+                separated_list1(
+                    many1(tag(" ")),
+                    many1(satisfy(|c| {
+                        ['0'..='9', 'a'..='z', 'A'..='Z', '-'..='-', '_'..='_']
+                            .iter()
+                            .any(|range| range.contains(&c))
+                    })),
```
mmarx, line 91: Shouldn't this be the same?
```rust
                many0(satisfy(|c| {
                    ['0'..='9', 'a'..='z', 'A'..='Z', '-'..='-', '_'..='_', ' '..=' ']
                        .iter()
                        .any(|range| range.contains(&c))
                })),
```

mmarx (changes requested):
The `expect` message needs fixing, the rest is more of a suggestion – especially `all_input_consumed`. Not sure if there is a reason for the more complex logic in `parse_bare_name`, but I don't see what we gain over just allowing spaces as part of the `satisfy`?

src/io/builder_proxy.rs
```diff
@@ -33,10 +36,20 @@ impl ColumnBuilderProxy for StringColumnBuilderProxy {
...
+            .unwrap_or(string.to_string());
```
ellmau, line 47:
```suggestion
        let parsed = super::parser::all_input_consumed(parse_dsv_constant)(string)
            .unwrap_or(string.to_string());
```

Checked the performance and it has been about 3 seconds faster on snomed to use `all_input_consumed` instead of the `map`-approach

src/io/dsv.rs
```diff
@@ -189,15 +189,15 @@ Boston;United States;4628910
                 .and_then(|u64| usize::try_from(u64).ok())
                 .and_then(|usize| dict.entry(usize))
                 .unwrap(),
-            "Boston"
+            "<Boston>"
```
ellmau, line 192: It might be a good test-case to leave one of the strings without the "< >" angles (which seems to work fine as well)

ellmau (changes requested):

ellmau:
Another thing which is missing, is to actually test the changes done in the Proxy builder done.

It does work to add the angled brackets in some or all input csv fields. Mixing is supported with the changes too.
I think this behaviour should be added as a test-case as well.

src/io/parser.rs
monsterkrampe: I think then bare names could have trailing spaces.

src/io/builder_proxy.rs
monsterkrampe: Wasn't aware of this one; will commit the suggestion by @ellmau below :)

src/io/builder_proxy.rs
```diff
@@ -7,6 +7,9 @@ use crate::{
         dictionary::Dictionary,
     },
 };
+
+use super::parser::parse_dsv_constant;
```
ellmau, line 11:
```suggestion
use super::parser::{all_input_consumed, parse_dsv_constant};
```

src/io/builder_proxy.rs
```diff
@@ -33,10 +36,16 @@ impl ColumnBuilderProxy for StringColumnBuilderProxy {
     generic_trait_impl!(VecT::U64);
     fn add(&mut self, string: &str, dictionary: Option<&mut Dict>) -> Result<(), Error> {
         self.commit();
+
+        // TODO: parsing of DSV and Program should be unified
+        // TODO: DSV parsing should depend on logical types
+        let parsed = super::parser::all_input_consumed(parse_dsv_constant)(string.trim())
+            .unwrap_or(string.to_string());
```
ellmau, line 43:
```suggestion
        let parsed = all_input_consumed(parse_dsv_constant)(string.trim())
            .unwrap_or(string.to_string());
```
Using import instead of fully qualifying the call.

ellmau (approved):
LGTM, suggestion to use imports instead of full qualification

mmarx (approved):

mmarx (approved):

PR #206, opened 2023-04-26
Better parser errors

mmarx:
Move to a custom error type for parser errors, and use `nom_locate` for position information. This produces much more informative parse errors. Fixes #172.

PR #207, opened 2023-04-26
Add support for @output directives

mmarx:
Add support for `@output` directives, allowing to narrow the IDB predicates that get written to disk. A CLI parameter `--write-all-idb-predicates` allows writing all IDB predicates for, e.g., debugging.

Based upon #206, so should not be merged before that.


ellmau (approved):

monsterkrampe (approved):
LGTM :)

PR #210, opened 2023-04-26
Provide Nemo as a package in the flake

mmarx:
Adds a `nemo` package to the flake, allowing, e.g., `nix run github:knowsys/nemo`.


mmarx:
Since we currently only have one crate that contains the binary, we might want to commit the top-level `Cargo.lock`, as is recommended for binaries. When we split out the library as a separate crate, it would get its own `Cargo.lock` that we can then ignore again.

ellmau (approved):
Looks good to me.

ellmau (approved):

PR #215, opened 2023-04-27
add support for signed integers in rle columns

matzemathics:


mmarx (approved):

src/physical/datatypes/storage_value.rs
```diff
@@ -24,30 +26,39 @@ pub enum StorageValueT {
 
 impl StorageValueT {
     /// Returns either [`Option<u32>`], answering whether the [`StorageValueT`] is of this datatype
-    pub fn as_u32(&self) -> Option<u32> {
+    pub fn get_u32(&self) -> Option<u32> {
```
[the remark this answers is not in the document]
mmarx: The more I think about it, I think it would be even better to get rid of these entirely and replace them with `TryFrom` implementations on the underlying types.

src/physical/tabular/operations/triescan_select.rs
```diff
@@ -126,12 +128,10 @@ impl<'a> TrieScanSelectEqual<'a> {
 }
 
 impl<'a> TrieScan<'a> for TrieScanSelectEqual<'a> {
-    #[allow(clippy::unnecessary_lazy_evaluations)] // not actually
-                                                   // unnecessary, as the subtraction might underflow
     fn up(&mut self) {
         debug_assert!(self.current_layer.is_some());
 
-        self.current_layer = self.current_layer.and_then(|index| index.checked_sub(1));
+        self.current_layer = self.current_layer.unwrap().checked_sub(1);
```
mmarx, line 134: What about `unwrap_or_default()` here, which would potentially avoid a panic?

tests/blackbox_integration.rs
```diff
@@ -97,6 +97,8 @@ impl TestCase {
 
         cmd.current_dir(self.test_dir.as_path())
             .arg("-s")
+            .arg("--log")
+            .arg("trace")
```
mmarx, line 101: Not sure we want this as a default.

src/physical/tabular/operations/triescan_select.rs
matzemathics: As per assertion above it should never panic.

tests/blackbox_integration.rs
matzemathics: I used this for debugging, because the test output was kinda useless. Would be nice if one could change this from the outside. Still the output is hidden, unless the test fails, so the only tradeoff here is a little slower execution time of the tests. I am also fine removing these though.

src/physical/datatypes/storage_value.rs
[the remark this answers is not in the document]
matzemathics: The reason was that I am confused when `as_` prefixed functions do not return a borrow (which is also stated in the API naming guidelines). Thinking about this some more though, the real confusion comes from the fact, that these do _two_ operations in one: Copy + TryInto. Implementing `TryFrom` is probably forbidden due to rusts orphaning rules (we do not own the trait nor the `u32` type).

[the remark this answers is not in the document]
matzemathics: How about
```suggestion
    pub fn try_into_u32(self) -> Option<u32> {
```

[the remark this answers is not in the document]
matzemathics: > Implementing TryFrom is probably forbidden due to rusts orphaning rules (we do not own the trait nor the u32 type).
Actually it is not. So I will go ahead and do this instead.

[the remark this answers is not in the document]
mmarx: I vaguely remember this being forbidden in old Rust versions, but it appears that since [Rust 1.41](https://blog.rust-lang.org/2020/01/30/Rust-1.41.0.html#relaxed-restrictions-when-implementing-traits), it's sufficient that the generic parameter is crate-local.

src/physical/tabular/operations/triescan_select.rs
mmarx: Then we can get rid of the assert and just use `expect(…)`, which will at least give us a nice error message, since we're doing the check anyways.

src/physical/datatypes/storage_value.rs
```diff
@@ -88,6 +52,27 @@ impl StorageValueT {
     }
 }
 
+macro_rules! storage_value_try_into {
+    ($variant:ident => $dst:ty) => {
+        impl TryFrom<StorageValueT> for $dst {
+            type Error = ();
```
mmarx, line 58: I think this is okay for now, we might want to make it a variant of `crate::Error::Error` at some point (possibly via an intermediate `StorageValueTryIntoError`, which could be a unit newtype).

mmarx (approved):

tests/blackbox_integration.rs
mmarx: The test cases should inherit `RUST_LOG` from the environment, so setting that to `trace` should work as well (haven't tested it, though).

matzemathics: Yes, I thought that too, but it does not work, because we use a constructor that ignores the environment.

mmarx: ah, all right then, let's leave this in

PR #216, opened 2023-04-27
Treat CSV input as basically being normalized

monsterkrampe:
I do not like this at all but the proper implementation is too slow.
I think some conceptual discussion is needed.
We should reconsider this once we support both `any` and `string` (but we cannot support both right now because the logical type information is not available to the CSV parser).


mmarx (approved):
Doesn't really look good to me, but fine for now ;)

Issue #86, opened 2023-01-05
Improve the Representation Format for Rules


aannleax:
Currently we use the same rule format for representing rules while parsing and also while planning their execution. This has two problems:

- The data structure contains too much detail that is necessary for parsing but superfluous later on
- The allowed syntax allows for more flexibility then is required. This results in a normalization step. It is then assumed in the rest of the code base that rules are in a normalized form. For example, we allow the user to type rules like `a(x, y) b(z, w), y = z -> c(x, w)` which we then translate to `a(x, y) b(y, w) -> c(x, w)`

The task is therefore to provide a separate data structure for rules for parsing and planning.

aannleax:
We have now had for some time the split between `Rule`s and `ChaseRule`s, which at least solves the problem outlined in this issue. A major rework of the rule model(s) is still needed, but I would suggest putting up a new issue for that with more concrete demands

PR #220, opened 2023-05-01
Stratified Negation

aannleax:
Implemented support for stratified negation. This involves:

- Adding the operators `ColumnScanSubtract` and `TrieScanSubtract` (probably not the best names). These can be viewed as generalizations of the `TrieScanMinus` operator that also allow to compute the difference between multiple tables of different arities
- Adding support for these iterators in the database
- Changing the logical model of rules such that it more clearly distinguishes between positive and negative body variables
- Change the `LabeledGraph` such that it now support edge labels and can compute a stratification which takes different kind of edge labels into account
- Implement a new stratified rule selection strategy


src/logical/model.rs
```diff
@@ -371,19 +371,41 @@ impl Filter {
...
+            }
+        }
+
+        (positive, negative)
+    }
+
     /// Construct a new rule.
-    pub fn new(head: Vec<Atom>, body: Vec<Literal>, filters: Vec<Filter>) -> Self {
+    pub fn new(head: Vec<Atom>, body: Vec<Literal>, positive_filters: Vec<Filter>) -> Self {
+        let (positive_body, negative_body) = Self::split_literal_vector(body);
```
monsterkrampe, line 401: I find it odd to have `split_literal_vector` as associated method on `Rule` when it is really independent of rules.
Maybe just inline what the method does right here.
Also, did you think about using [`partition`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.partition)?

src/logical/execution/planning/negation.rs
```diff
@@ -0,0 +1,142 @@
...
+        tabular::operations::{triescan_minus::SubtractInfo, triescan_project::ProjectReordering},
+        util::mapping::traits::NatMapping,
+    },
+};
+
+struct AtomInfo {
+    restricted_variable_order: VariableOrder,
+    reorder: ColumnOrder,
+    projection: ProjectReordering,
+    subtraction: SubtractInfo,
```
monsterkrampe, line 25: More of a general comment:
I think also in other places many helper structs are named in a way that does not convey much information of what is stored in them. For instance, `AtomInfo` could basically be anything. On the other hand `SubtractInfo` at least gives a hint on what this data is used for but the contents are also not clear from the name.
At least for me that makes understanding what is supposed to happen in the code quite hard.

src/physical/tabular/operations/triescan_minus.rs
```diff
@@ -1,14 +1,260 @@
...
+                    .used_layers
+                    .iter()
+                    .max()
+                    .expect("There should be at least one layer that is used.")
+            })
+            .collect();
+        let mut column_scans = Vec::<UnsafeCell<ColumnScanT<'a>>>::with_capacity(arity_main);
+
+        for layer in 0..arity_main {
+            unsafe {
```
monsterkrampe, line 105: Can we limit the scope of the `unsafe` for example by having an additional `let` expression before the `if`?

src/physical/util/labeled_graph.rs
```diff
@@ -16,23 +16,25 @@ use petgraph::{EdgeType, Graph};
 /// A [`NodeIndex`] is invalidated once a node is removed from the graph,
 /// hence the interface only permits adding new nodes.
 #[derive(Debug)]
-pub struct NodeLabeledGraph<Label, Type>
+pub struct LabeledGraph<NodeLabel, EdgeLabel, Type>
```
monsterkrampe, line 19: Isn't this mostly or even exclusively used in the logical layer? (I'm wondering why it is defined in `physical/util`.)

monsterkrampe (changes requested):
Some small-ish things. Looks good otherwise :)

src/logical/execution/selection_strategy/strategy.rs
```diff
@@ -1,12 +1,27 @@
...
+use thiserror::Error;
+
 use crate::logical::{model::Rule, program_analysis::analysis::RuleAnalysis};
 
+/// Errors that can occur while creating a strategy.
+#[derive(Error, Debug, Copy, Clone)]
+pub enum SelectionStrategyError {
+    /// Non-Stratifyable
+    #[error("The rules of the program cannot be stratified.")]
+    NonStratifiableProgram,
```
mmarx, line 12: The wording suggests that we're talking about something like effectively stratifiable programs – but we're really just supporting stratified programs. I would suggest to have “is not stratified” and `NonStratifiedProgram`.

src/logical/model.rs
```diff
@@ -513,26 +531,50 @@ impl Rule {
 
     /// Return the body literals of the rule - immutable.
     #[must_use]
-    pub fn body(&self) -> &Vec<Literal> {
-        &self.body
+    pub fn positive_body(&self) -> &Vec<Atom> {
+        &self.positive_body
+    }
```
mmarx, line 536: I've seen several `positive_body().chain(….negative_body())` below – why not add something like this?
```rust
pub fn body(&self) -> impl Iterator<Item = Atom> {
  self.positive_body.iter().chain(self.negative_body.iter())
}
```

src/logical/program_analysis/analysis.rs
```diff
@@ -480,21 +498,11 @@ impl Program {
...
-                // check for negation
-                if matches!(atom, Literal::Negative(_)) {
-                    return Err(RuleAnalysisError::UnsupportedFeatureNegation);
-                }
-
-                // check for consistent predicate arities
-                let arity = atom.terms().len();
-                if arity != *arities.entry(atom.predicate()).or_insert(arity) {
-                    return Err(RuleAnalysisError::UnsupportedFeaturePredicateOverloading);
-                }
```
mmarx, line 493: Why drop this? What code replaces this check?

src/logical/program_analysis/analysis.rs
```diff
@@ -533,7 +541,11 @@ impl Program {
         }
 
         for (rule, analysis) in self.rules().iter().zip(analyses.iter()) {
-            for filter in rule.filters() {
+            for filter in rule
+                .positive_filters()
+                .iter()
+                .chain(rule.negative_filters())
```
mmarx, line 547: Again, would be useful to have `Rule::filters` which does the chaining transparently.

src/logical/model.rs
```diff
@@ -371,19 +371,41 @@ impl Filter {
...
-    body: Vec<Literal>,
+    /// Positive Body literals of the rule
+    positive_body: Vec<Atom>,
     /// Filters applied to the body
-    filters: Vec<Filter>,
+    positive_filters: Vec<Filter>,
+    /// Negative Body literals of the rule
+    negative_body: Vec<Atom>,
+    /// Filters applied to the body
+    negative_filters: Vec<Filter>,
```
mmarx, line 381: This greatly enlarges the class of rules that we will not be able to serialise back into their original syntactic representation, as we lose the relative order of positive and negative literals. Instead, we should really push #86.

src/physical/columnar/operations/columnscan_minus.rs
```diff
@@ -214,10 +214,186 @@ where
...
+    T: 'a + ColumnDataType,
+{
+    type Item = T;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.current_value = if let Some(next_value) = self.scan_main.next() {
+            self.move_follow_scans(next_value)
+        } else {
+            None
+        };
```
mmarx, line 354:
```rust
self.current_value = self.scan_main.next().and_then(|next_value| self.move_follow_scans(next_value));
```

src/physical/columnar/operations/columnscan_minus.rs
```diff
@@ -214,10 +214,186 @@ where
...
+impl<'a, T: Ord + Copy + Debug> ColumnScan for ColumnScanSubtract<'a, T>
+where
+    T: 'a + ColumnDataType,
+{
+    fn seek(&mut self, value: T) -> Option<T> {
+        self.current_value = if let Some(next_value) = self.scan_main.seek(value) {
+            self.move_follow_scans(next_value)
+        } else {
+            None
+        };
```
mmarx, line 369:
```rust
self.current_value = self.scan_main.seek(value).and_then(|next_value| self.move_follow_scans(next_value));
```

src/physical/management/execution_plan.rs
```diff
@@ -128,6 +137,8 @@ pub enum ExecutionOperation {
     AppendColumns(ExecutionNodeRef, Vec<Vec<AppendInstruction>>),
     /// Append (the given number of) columns containing fresh nulls.
     AppendNulls(ExecutionNodeRef, usize),
+    /// Operation wich subtracts a list multiple table potentially of different arities from another table.
```
mmarx, line 140: “a list multiple table” – there's something missing here.

src/physical/management/execution_plan.rs
```diff
@@ -815,6 +888,16 @@ impl ExecutionTree {
                 assert!(permutation.is_identity());
                 Self::satisfy_leapfrog_recurisve(subnode.clone(), permutation.clone());
             }
+            ExecutionOperation::Subtract(subnode_main, subnodes_subtract, _) => {
+                // TODO: A few other changes are needed to make this a bit simpler.
+                // Will update it then
```
mmarx, line 893: What are the changes?

src/physical/util/labeled_graph.rs
```diff
@@ -76,3 +172,58 @@ where
...
+    use super::LabeledGraph;
+
+    #[derive(Debug, Eq, PartialEq, Hash, Clone)]
+    enum EdgeLabel {
+        Positive,
+        Negative,
+    }
+
+    #[test]
+    fn test_stratify() {
```
mmarx, line 189: Tests shouldn't start with `test_`, they already have `test::` in their name.

src/physical/util/labeled_graph.rs
```diff
@@ -76,3 +172,58 @@ where
...
+
+        assert!(stratums.is_some());
+        assert_eq!(
+            stratums.unwrap(),
+            vec![vec![node_a], vec![node_b, node_c], vec![node_d]]
+        );
+    }
+
+    #[test]
+    fn test_non_stratifiable() {
```
mmarx, line 217: See above. Also should really be called `non_stratified`.

src/physical/tabular/operations/triescan_minus.rs
```diff
@@ -329,4 +592,226 @@ mod test {
         assert_eq!(diff_next(&mut diff_iter), None);
         assert_eq!(diff_current(&mut diff_iter), None);
     }
+
+    #[test]
+    fn test_subtract() {
```
mmarx, line 597: See below, no `test_` in test names.

src/physical/tabular/operations/triescan_minus.rs
```diff
@@ -1,14 +1,260 @@
...
+                        subtract_indices,
+                        follow_indices,
+                    ));
+
+                    column_scans.push(UnsafeCell::new(ColumnScanT::U64(ColumnScanCell::new(
+                        new_scan,
+                    ))));
+                }
+            }
+        }
```
mmarx, line 148: Doesn't even require any additional `let`s, you can just push the `unsafe` inside: 
```rust
        for layer in 0..arity_main {
            if let ColumnScanT::U64(left_scan_enum) =
                &unsafe { *trie_main.get_scan(layer).unwrap().get() }
            {
                let mut scans_follower = Vec::<Option<&ColumnScanCell<'a, u64>>>::new();

                let mut subtract_indices = Vec::new();
                let mut follow_indices = Vec::new();

                for (subtract_index, (trie_subtract, info)) in
                    tries_subtract.iter().zip(infos.iter()).enumerate()
                {
                    let used_layer = info.used_layers.iter().position(|&l| l == layer);
                    let is_last = layer == last_used_layers[subtract_index];

                    if let Some(used_layer) = used_layer {
                        if let ColumnScanT::U64(subtract_scan_enum) =
                            &unsafe { *trie_subtract.get_scan(used_layer).unwrap().get() }
                        {
                            scans_follower.push(Some(subtract_scan_enum));
                        }

                        if is_last {
                            subtract_indices.push(subtract_index);
                        } else {
                            follow_indices.push(subtract_index);
                        }
                    } else {
                        scans_follower.push(None);
                    }
                }

                let new_scan = ColumnScanEnum::ColumnScanSubtract(ColumnScanSubtract::new(
                    left_scan_enum,
                    scans_follower,
                    subtract_indices,
                    follow_indices,
                ));

                column_scans.push(UnsafeCell::new(ColumnScanT::U64(ColumnScanCell::new(
                    new_scan,
                ))));
            }
        }
```

mmarx (changes requested):
Some nitpicks and some small improvements, but we should really prioritise #86 over pushing the logical data model even further from what we need there.

src/logical/model.rs
aannleax: Partition is not that convenient since I get two vectors of literals instead of two vectors of atoms. I went with the inline solution.

src/logical/execution/planning/negation.rs
aannleax: Well the thing is that those really are just "random" collections of objects that just need to be used together. As such, the overall struct does not represent any kind of concept that can have a meaningful name. 

I changed `AtomInfo` to `AtomNegationInfo` and added a bit more comments. But I don't think its going to get better than that.

src/logical/program_analysis/analysis.rs
aannleax: Moved to the for loop below

src/logical/model.rs
aannleax: Yes this is of course temporary until #86 get done.

src/physical/management/execution_plan.rs
aannleax: What is needed here is a way to change this operation node such that it outputs columns in a different order according to the permutation. 

I'm too lazy now to think this through  for this operation (and the others that have a similar comment). Also, I most likely will rework how this works entirely anyways, so it might end up not mattering in the end.

src/logical/model.rs
monsterkrampe: I see. Looks fine :)

src/logical/execution/planning/negation.rs
aannleax: Once we have a separate rule representation from which execution plans can be built more easily, much of this code will change anyways, so I wouldn't worry about it too much

monsterkrampe: Sounds good :)

monsterkrampe (approved):
From my point of view this can be merged now. It seems like at lot of things will change again though once #86 has been handled in some way.

aannleax:
I merged the changes from #233 into this PR

mmarx (changes requested):
Some of my comments from the last review seem to still be unaddressed – at the very least, the broken grammar in the doc comment still needs fixing.

aannleax:
> Some of my comments from the last review seem to still be unaddressed – at the very least, the broken grammar in the doc comment still needs fixing.

Ups github hid them; should now be included

mmarx (approved):

PR #222, opened 2023-05-05
Decouple csv reading from physical layer

monsterkrampe:


monsterkrampe:
@ellmau you can take over from here.
Everything seems to work as intended. I'm not sure if all the lifetime madness is really necessary; maybe this can be simplified.
Also, the things from `io` should probably distributed to the physical and the logical layer. I think this should be possible with the current implementations.
I also did not check the performance of CSV reading.

ellmau:
> As a follow up step, one could bring back the proper term parsing for the `Any` type and introduce an additional `String` type. But I think that this is out of scope of this PR.

I think so too, we should try to have smaller PRs anyway.

src/logical/execution/execution_engine.rs
```diff
@@ -111,12 +111,19 @@ impl<Strategy: RuleSelectionStrategy> ExecutionEngine<Strategy> {
...
-                DataSource::DsvFile { file, delimiter } => TableSource::DSV {
-                    file: *file.clone(),
-                    delimiter: *delimiter,
-                },
+                DataSource::DsvFile { file, delimiter } => {
+                    let logical_types = analysis
+                        .predicate_types
+                        .get(predicate)
+                        .cloned()
+                        .unwrap_or_else(|| (0..arity).map(|_| LogicalTypeEnum::Any).collect());
```
mmarx, line 121:
```rust
                    .unwrap_or_else(|| vec![LogicalTypeEnum::Any; *arity]),
```

mmarx (changes requested):
Looks good, just a small simplification

mmarx (approved):

Issue #227, opened 2023-05-15
Split rule model and intermediate representation(s)


mmarx:
Split the user-facing rule model from the intermediate representation used for the chase implementation (and possibly from an intermediate representation used for parsing and type inference). Sets the basis for #86, which in turn enables #220 without sacrificing parsing faithfulness.

monsterkrampe:
This should be done by now.

PR #233, opened 2023-05-16
Refactor (split) rule model

mmarx:
Work towards #86 and #227. While this is a draft PR, it might frequently be rebased or otherwise force-pushed, please don't base other code on this yet.


mmarx:
This does the bare minimum of splitting out a chase variant of the rule model, with the adapted versions of `Program` and `Rule` from #220, and `TryFrom` implementations to convert from the canonical rule model.

aannleax (approved):

PR #235, opened 2023-05-17
insert TrieSerializer trait, do writing in logical layer

matzemathics:


mmarx (dismissed):
One small nit to pick, looks good otherwise!

Issue #228, opened 2023-05-15
Rust API for reasoning tasks


ellmau:
To use the nemo library for reasoning tasks, a rudimentary API would be needed.

# API
A MVP for this includes the following functionality. Note that this is the absolute minimum and more features can (and should be) added later.

## load rule file (parts)
This shall take a given string and parse it like any given input-program.
In contrast to the current model, where the whole program is loaded in one step,
incremental calls with different rules shall be possible.
A parse-error should be reported back and only this call of the load-directive shall be rolled back.
## reasoning
Calling the actual reasoning process, which includes reading of external resources and creating the full reasoning closure of the given program.
## query
Query some results from the reasoning process and provide ways to view the results.
This can be an object model representation of the result, as well as direct access to the efficient writing methods, already available (i.e. csv output writer)
## reset
Reset the state of the reasoning engine, where no rules have been loaded and no conclusions have been drawn.

# Architecture
It will be necessary to track the state of the input and the parsed rule model. A way to roll back/commit changes to the already parsed rules need to be introduced as well.

mmarx:
Landed as part of #236.

PR #236, opened 2023-05-22
Feature/228 rust high level api

ellmau:
First suggestion on a high level api with few methods, as mentioned in #228

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -1,0 +1,3 @@
+//! api
+
+pub mod high_level;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -673,6 +673,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "dashmap"
+version = "5.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "907076dfda823b0b36d2a1bb5f90c96660a5bbcd7729e10727f07858f22c4edc"
+dependencies = [
+ "cfg-if",
+ "hashbrown",
+ "lock_api",
+ "once_cell",
+ "parking_lot_core",
+]
+
+[[package]]
 name = "derive_arbitrary"
 version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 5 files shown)


matzemathics:
As I already explained in the last meeting, I don't like the concept of a stateful control object, nor do I see any benefit in creating one (apart from maybe reducing the number of opaque types from 2 to 1).

The simplest solution, as I see it is to have three functions, two opaque types and no hidden state:
- `load` takes a path and returns an `ExecutionEngine` (wrapped in an error)
- `reason` takes an `ExecutionEngine` and returns a Vec of TableSerializers
- `write` takes the Vec of TableSerializers and writes the result into a directory

This way, there is no state machine and thus no API errors.

src/api.rs
```diff
@@ -0,0 +1,152 @@
...
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum Error {
+    /// nemo error
+    #[error(transparent)]
+    Nemo(#[from] crate::error::Error),
+    /// API error
+    #[error("API error {0} occurred: {1}")]
+    API(usize, String),
```
mmarx, line 32: Just no. You're re-inventing an `enum`, badly, with two different fields that can even drift separately. All of these should rather be variants in an enum, with the error message associated to the variant, and, should we really need an error code at all (what for? this is a rust API, after all, callers will be doing `match`es on the variants, and foreign language bindings are explicitly not in scope for this API), this can be done as a method on the error enum.

src/api.rs
```diff
@@ -0,0 +1,152 @@
...
+mod test {
+    use std::fs::read_to_string;
+
+    use super::*;
+    use assert_fs::TempDir;
+    use test_log::test;
+
+    #[test]
+    fn reason() {
+        let mut nemo_ctl = NemoControl::default();
```
mmarx, line 126: should be `NemoControl::new()` (and that should be implemented).

src/api.rs
```diff
@@ -0,0 +1,152 @@
...
+        let bytes = include_bytes!("../resources/testcases/lcs-diff-computation/run-lcs-10.rls");
+        let input = String::from_utf8(bytes.to_vec()).unwrap();
+
+        let temp_dir = TempDir::new().unwrap();
+
+        assert!(nemo_ctl.load(input).is_ok());
+        log::info!("changed to {:?}", std::env::current_dir().unwrap());
```
mmarx, line 133: Not actually changed anything, though, so the log message is misleading.

ellmau:
> As I already explained in the last meeting, I don't like the concept of a stateful control object, nor do I see any benefit in creating one (apart from maybe reducing the number of opaque types from 2 to 1).
> 
> The simplest solution, as I see it is to have three functions, two opaque types and no hidden state:
> 
>     * `load` takes a path and returns an `ExecutionEngine` (wrapped in an error)
> 
>     * `reason` takes an `ExecutionEngine` and returns a Vec of TableSerializers
> 
>     * `write` takes the Vec of TableSerializers and writes the result into a directory
> 
> 
> This way, there is no state machine and thus no API errors.

In the conversation last week, we said that there might be different API implementations and that this easy high level one, which is similar to how it works in rulewerk should be done first, to allow the students at the student projects for a very basic and easy usage of nemo in a similar style to how the rulework commands for loading, reasoning, materializing, and resetting works.

mmarx:
So I wasn't present for this conversation, but given the choice between an opaque handler object with several methods that I can call in multiple wrong orders and three freestanding methods with argument types that force me to call them in the right order, it's _very_ clear to me which one is easier to use, and it's not the one with the handler object.

src/api.rs
```diff
@@ -106,43 +109,47 @@ impl NemoControl {
...
+    use serial_test::serial;
     use test_log::test;
 
+    use super::*;
+
     #[cfg_attr(miri, ignore)]
     #[test]
-    fn reason() {
+    #[serial]
+    fn reason2() {
```
mmarx, line 124: What is this test for? It seems you are testing the same code twice? These tests should make sure the API works as expected, not do full-scale integration tests (we do those elsewhere already). In particular, what there _should_ be tests for (actually not, because we really rather should get rid of the state object) is that, e.g., calling API methods in the wrong order returns _the expected errors_.

mmarx (changes requested):

mmarx:
Another benefit of getting rid of the state object is that we can then properly unit test the API, which is currently impossible.

mmarx:
Thanks for the rebase, all of the comments from above still apply, although they are now outdated since you renamed the file.

mmarx:
I'm also very much opposed to calling this `api::high_level`, since that is relative to whichever APIs we might also implement (think `api::higher_lewel`, `api::even_higher_level`, `api::burj_khalifa`).

ellmau:
> I'm also very much opposed to calling this `api::high_level`, since that is relative to whichever APIs we might also implement (think `api::higher_lewel`, `api::even_higher_level`, `api::burj_khalifa`).

I'm fine with every kind of naming for this API. I don't think it will be used much as soon as the opaque typed variant is implemented.

ellmau:
Overall I'll let this option as a draft pr and will try to get another option done in the following days.

matzemathics:
I think this can be closed in favour of #237. Feel free to reopen if you wanted to follow up on this.

PR #237, opened 2023-05-23
API with opaque types

ellmau:
Here is the second proposal for an API.

In my opinion that one is easier to use and can be utilized and extended easily.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -95,11 +95,7 @@ pub fn reason(engine: &mut Engine) -> Result<Vec<Predicate>, Error> {
     let results = engine.0.combine_results()?;
     Ok(results
         .into_iter()
-        .filter(|(_id, table_id)| table_id.is_some())
-        .map(|(id, table_id)| Predicate {
-            name: id,
-            table_id: table_id.expect("filtered all None values"),
-        })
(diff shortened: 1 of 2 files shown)


src/api.rs
```diff
@@ -0,0 +1,148 @@
...
+///
+/// # Note
+/// If there have been `@source` routines in the parsed rules, all relative paths are resolved with the current working directory
+pub fn reason(engine: &mut Engine) -> Result<Vec<Predicate>, Error> {
+    engine.0.execute()?;
+    let results = engine.0.combine_results()?;
+    Ok(results
+        .into_iter()
+        .filter(|(_id, table_id)| table_id.is_some())
+        .map(|(id, table_id)| Predicate {
```
matzemathics, line 99: Maybe `filter_map` instead of `filter` + `map`

matzemathics (changes requested):
One small nit, but overall I like this much better.

mmarx (changes requested):
Same nit as @matzemathics mentioned, looks good otherwise!

mmarx (approved):

matzemathics (approved):

PR #239, opened 2023-05-23
Build support for Web Assembly targets

rlwww:


src/meta/timing.rs
```diff
@@ -145,51 +159,63 @@ impl TimedCode {
...
-        self.info.start_system = None;
-        self.info.start_process = None;
-        self.info.start_thread = None;
-        self.info.runs += 1;
-
-        duration_thread
+        #[cfg(not(feature = "timing"))]
+        {
+            // This is not reachable, but the function would otherwise need a return value
+            panic!();
```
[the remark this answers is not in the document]
matzemathics: Ah, I see, it is not unreachable, but won't be called if timing is inactive. In this case I would just propose to make the comment the error message of `panic!`.
Also saying it is not reachable is sort of a lie, because it depends on the outside code. Maybe state that the operation is not supported?

src/meta/timing.rs
```diff
@@ -145,51 +159,63 @@ impl TimedCode {
 
     /// Start the next measurement
     pub fn start(&mut self) {
-        if cfg!(test) {
+        if cfg!(test) || cfg!(not(feature = "timing")) {
```
matzemathics, line 162: This could be a compile-time check instead of a runtime check. Maybe it is nicer to provide a second implementation?
```rust
    #[cfg(any(feature = "test", not("feature = timing")))]
    pub fn start(&mut self) {
         // No-Op 
    }
```

matzemathics (commented):
Looks good, just some ideas.

src/meta/timing.rs
```diff
@@ -120,7 +138,7 @@ impl TimedCode {
 
     /// Navigate to a subblock (use forward slash to go multiple layers at once)
     pub fn sub(&mut self, name: &str) -> &mut TimedCode {
-        if cfg!(test) {
+        if cfg!(test) || cfg!(not(feature = "timing")) {
```
matzemathics, line 141: Maybe do the same thing here?

mmarx (approved):

matzemathics (approved):
Looks fine, but maybe its nicer to have all functions implemented in the same style

PR #240, opened 2023-05-23
Decentralized api

matzemathics:
As an alternative to #237, here is a refactor which removes the "api" module, but still incorporates some ideas from the API design. This is because it's somewhat unidiomatic to have a module dedicated to "API", because everything that is public is already part of our API.

I also based a python binding implementation on this, see https://github.com/matzemathics/nmo-python/tree/main


mmarx (changes requested):
I like the general idea, especially having less code duplication between “API” and “CLI app”. 

However, having the very-high-level API that abstracts away nearly everything is _also_ helpful (and can certainly be implemented in terms of the new primitives). 

Also, just a few small comments that definitely need clarification.

src/cli.rs
```diff
@@ -0,0 +1,128 @@
...
+    /// Initialising Logging
+    ///
+    /// Sets the logging verbosity to the given log-level in the following order:
+    ///  * `Info`, `Debug`, `Trace`; depending on the count of `-v`
+    ///  * `Error` when `-q` is used
+    ///  * The `NMO_LOG` environment variable value
+    ///  * `Warn` otherwise
+    pub fn initialize_logging(&self) {
+        let mut builder = env_logger::Builder::new();
+        builder.parse_env("NMO_LOG");
```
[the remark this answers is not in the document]
matzemathics: Yeah, reading from the environment was broken previously, because the cli arguments always overwrote it (even if there were none). While fixing this I also changed the name, just because the environment argument does not control anything rust specific but can be seen as an extension of nemo's cli interface.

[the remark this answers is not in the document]
matzemathics: I can take care of the scripts or revert, but would probably prefer `NMO_LOG`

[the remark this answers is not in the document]
mmarx: I think it was actually working as intended – certainly working as documented, though. In general, I would like to be able to override the environment setting with the command line arguments. Regarding “anything rust specific”, it certainly _does_ also control logging in crates that are not our own (but we might still want to rename it).

src/logical/execution/execution_engine.rs
```diff
@@ -260,7 +287,19 @@ impl<Strategy: RuleSelectionStrategy> ExecutionEngine<Strategy> {
             result_ids.push((predicate.clone(), table_id));
         }
 
-        Ok(result_ids)
+        Ok(result_ids
+            .into_iter()
+            .map(|(p, id)| IdbPredicate {
+                identifier: p,
+                table_id: id,
+            })
```
mmarx, line 295:
```rust
.map(|(identifier, table_id)| IdbPredicate { identifier, table_id })
```

src/error.rs
```diff
@@ -41,6 +43,12 @@ pub enum Error {
     /// Error when converting floating type to integer point value
     #[error("Floating type could not be converted to integer value")]
     FloatingPointToInteger,
+    /// Error if no input rule files are specified
+    #[error("No inputs were specified")]
+    NoInput,
+    /// Error if the user asked for an unimplemented feature
+    #[error("The specified input uses {0}, which is an unimplemented feature")]
+    Unimplemented(String),
```
mmarx, line 51: Please don't do stringly typed errors, have one variant for each feature (there are already some variants for unsupported features). Also, didn't we support multiple input files before?

src/io/parser.rs
```diff
@@ -26,6 +26,18 @@ pub(crate) mod sparql;
 pub(crate) mod turtle;
 pub use types::{LocatedParseError, ParseError, ParseResult};
 
+/// Parse a program in the given `input`-String and return an [`Program`].
```
mmarx, line 29: `an` ~> `a`

mmarx (changes requested):

src/cli.rs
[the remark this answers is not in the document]
matzemathics: > it certainly does also control logging in crates that are not our own

ah, I was not aware of that.

Previously when self.verbose was zero (i.e. no verbose flag) and no other cli parameters were set, then the log filter would be set to `Warn`, hence why I added the if > 0.

That was also the reason for me to add the trace parameter in the blackbox tests, which is now removed, because it slowed down the tests significantly.

src/error.rs
matzemathics: Yes, I was not sure if I should leave it this way, because it was just concatenating multiple files into a single string. Maybe this should just be left unimplemented until we get around to modules, but I can certainly also recreate the old behaviour.

matzemathics: One drawback with the variant for each feature approach is that implementing a feature and removing that variant is a breaking change. But who cares anyway

mmarx:
> However, having the very-high-level API that abstracts away nearly everything is also helpful (and can certainly be implemented in terms of the new primitives).

Still not sure if this is really what we want, let's discuss this tomorrow in the meeting?

src/api.rs
```diff
@@ -125,12 +96,12 @@ mod test {
...
     fn reason() {
         let mut engine =
-            load("./resources/testcases/lcs-diff-computation/run-lcs-10.rls".to_string()).unwrap();
+            load("./resources/testcases/lcs-diff-computation/run-lcs-10.rls".into()).unwrap();
         let cur_dir = std::env::current_dir().unwrap();
         std::env::set_current_dir("./resources/testcases/lcs-diff-computation/").unwrap();
         let results = super::reason(&mut engine).unwrap();
         std::env::set_current_dir(cur_dir).unwrap();
-        assert_eq!(results.len(), 20); // 2 empty tables, which are omitted
+        assert_eq!(results.len(), 22); // 2 empty tables, which are _not_ omitted
```
ellmau, line 104: That comment is not very informative now.
The previous version removed all IDBPredicates without table_ids from the result, so it was reasonable to mention that 2 empty tables are omitted.
As this filtering is no longer in place, it is not needed to mention that the two empty tables are not omitted.

src/cli.rs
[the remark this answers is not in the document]
mmarx: > Previously when self.verbose was zero (i.e. no verbose flag) and no other cli parameters were set, then the log filter would be set to Warn, hence why I added the if > 0.

I agree, that should not happen.

mmarx (approved):

PR #229, opened 2023-05-15
seperate TrieScan and PartialTrieScan traits, rewrite materialize

matzemathics:
The next step is to implement materialize in terms of the new `TrieScan` trait, which should be implemented for `TrieScanPrune` and `TrieScanGeneric`.

Furthermore the plan generation needs to be updated to have all materialized nodes implement the `TrieScan` trait.

PR #241, opened 2023-05-24
Combine efficiency

matzemathics:
Fixes an inefficiency were currently, every time combine is called, a union of all single tables is computed. In particular, calling `combine_tables` twice computes the same table twice and appends it twice to the sub-table handler.


mmarx (commented):
Not a full review yet, just a minor comment

mmarx (approved):
Looks good, can you please rebase this on `main` and/or #220?

matzemathics:
Note that this is based on #229 , which should either be merged first or closed in favour of this.

Issue #85, opened 2023-01-05
Design and implement type system in logical layer


aannleax:
As part of #70, we planned to have a separate type system on the physical layer and the logical layer. While the former deals with low-level types like u32 or floats, the latter deals with more abstract concepts like "String", "Integer" and so on. The task is to implement this functionality.

Some more considerations:
- This would require to have a translation from the logical type into a `TrieSchemaElement` which currently represents the type in the physical layer
- Some type analysis on the program level is necessary. E.g. Joins can only be performed on columns of the same logical type. This would require an implementation of type propagation through a predicate position graph
- One should think about having a uniform system of representing values. Currently the CSV-loading code for example is independent from the rule parsing code.
- Type declarations should be part of the rule language: For example in the @source statements
- Users should be allowed to not specify a type. This would then be automatically deduced while loading the files. This could be tricky because of the type analysis that needs to be performed (E.g. one might find that constraints of the type analysis are violated while loading the file). 
- Users should be allowed to overspecify a type. E.g. a user might say that a column should be of type `Integer(32)` even though we might not differentiate between integers of different sizes in the logical layer

monsterkrampe:
Related discussion: #185

monsterkrampe:
A basic type system with `Any`, `String`, `Integer` and `Float64` has been implemented.
Since there is no type casting when reasoning yet, only `Any` and `String` are compatible because they happen to have the same physical representation.

Still, since this marks a basic milestone of the type system development, I hereby close this issue.
We should open new issues for open points and use the existing discussion for clarifying additional questions.

PR #243, opened 2023-05-25
Add logical string type

monsterkrampe:
addresses #85

This PR implements a logical `String` type that just pipes through every string to the physical layer without doing any sophisticated parsing. In contrast, the `Any` type now does exactly that, i.e. parses its input string as RDF term, which is in turn mapped appropriately to the physical layer.

If no type declarations are specified, all columns from CSV inputs are treated as `String`. (Every Datasource variant has its own default type; for RDF files this would be `Any`.)
Note that the String `"A"` and Constants the constant `A` look the same in CSV: `A`
Further note that the physical representation of the logical `String` `"A"` (or `"A"^^<http://www.w3.org/2001/XMLSchema#string>` or `"A"@cz`) and the logical `Any` constant `A` are also the same: `A`.
String literals `"A"` or (`"A"^^<http://www.w3.org/2001/XMLSchema#string>` or `"A"@cz`) in columns with a logical type of `Any` are represented as `"A"`, both in the physical layer and when written to CSV.

The normalization of rdf terms like `"A"^^<http://www.w3.org/2001/XMLSchema#string>` for the `String` type only applies to terms from the rule file. `"A"^^<http://www.w3.org/2001/XMLSchema#string>` in a `String` column of a CSV is treated as the `String` `"\"A\"^^<http://www.w3.org/2001/XMLSchema#string>"`.

I think the confusion has now been completed :rocket: 
Goes without saying but please don't be shy to raise concerns :)


monsterkrampe:
!!! NOTE: I did not check if any of our usual benchmarks break here !!!

resources/testcases/datalog_constants/sources/dataA.csv
```diff
@@ -1,4 +1,4 @@
-<A>,<B>,<C>
-<B>,<B>,<D>
-<A>,<E>,<F>
-<C>,<D>,<D>
\ No newline at end of file
+A,B,C
+B,B,D
+A,E,F
+C,D,D
```
monsterkrampe, line 4: These are all `String`s here.

resources/testcases/restricted_block/run.rls
```diff
@@ -1,4 +1,6 @@
+@declare data(any, any).
```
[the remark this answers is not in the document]
aannleax: Shouldn't the type inference be able to detect this though? I suppose the problem is that "String" is assigned too early as the datatype for the loaded predicate

src/io/dsv.rs
```diff
@@ -309,35 +309,35 @@ Boston;United States;4628910
...
+The next 2 columns are empty;;;789
 ";
 
         let expected_result = [
-            ("Boston", "United States", 4628910),
-            ("Dresden", "Germany", 1234567),
-            ("My Home Town", "\"Some<where >Nice\"", 2),
+            ("Boston", "United States", "Some String", 4628910),
+            ("Dresden", "Germany", "Another String", 1234567),
+            ("My Home Town", "\"Some<where >Nice\"", "<https://string.parsing.should/not/change#that>", 2),
```
mmarx, line 327: I think using a raw string literal might make this more readable:

```rust
("My Home Town", r#""Some<where >Nice""#, "<https://string.parsing.should/not/change#that>", 2),
```

src/io/dsv.rs
```diff
@@ -309,35 +309,35 @@ Boston;United States;4628910
...
             (
                 "Trailing Spaces do not belong to the name",
                 "What about spaces in the beginning though",
+                "  what happens to spaces in string parsing?  ",
                 123,
             ),
             (
                 "\"Do String literals work?\"",
                 "\"Even with datatype annotation?\"",
+                "\"even string literals should just be piped through\"^^<http://www.w3.org/2001/XMLSchema#string>",
```
mmarx, line 337: see above

src/io/dsv.rs
```diff
@@ -309,35 +309,35 @@ Boston;United States;4628910
...
                 456,
             ),
-            ("The next column is empty", "\"\"", 789),
+            ("The next 2 columns are empty", "\"\"", "", 789),
```
mmarx, line 340: see above

mmarx (approved):
Looks good to me, with the possible improvement of using raw string literals to make the test case more readable.

resources/testcases/restricted_block/run.rls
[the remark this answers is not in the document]
aannleax: I wonder if the way to fix this is to say something like: 

- Every explicit type declaration triggers type inferences (by that I mean "is a starting point in the dfs-traversel of the predicate-position graph for inferring types)
- Every position which contains an existential variable triggers the type inference
- After this, every unknown type will be replaced by defaults, which will be `any` for idb or the `source.default_type()` for edbs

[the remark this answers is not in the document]
monsterkrampe: I can give this some thought again.
I definitely wanted to propagate the default source types since otherwise one gets a lot of type clashes as well.
So at the moment the type propagation is done based on the explicit declarations and the data source default types.

aannleax (changes requested):
Looks very good. 

My only problem is the type inference as mentioned above.

resources/testcases/restricted_block/run.rls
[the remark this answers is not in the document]
aannleax: Yeah you're right; non-any default types would have to be propagated again ... 
Might not be so simple then

aannleax (approved):

src/io/dsv.rs
monsterkrampe: Good idea, thanks!

resources/testcases/restricted_block/run.rls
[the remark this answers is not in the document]
monsterkrampe: The existential positions are also propagated now! 
(Going to test with our typical examples/benchmarks.)

src/io/parser/types.rs
```diff
@@ -100,6 +100,9 @@ pub enum ParseError {
         r#"SPARQL data source for predicate "{0}" has arity {1}, but {2} variables are given"#
     )]
     SparqlSourceInvalidArity(String, usize, usize),
+    /// A data source declarations types do not match its arity.
+    #[error(r#"Data source for predicate "{0}" has arity "{1}" but {2} types are given."#)]
+    SourceInvalidTypeArity(String, usize, usize),
```
mmarx, line 105: `declaration's`, but we should rather make this something that cannot occur, by only allowing either of arity and types to be given, not both.

src/logical/builder_proxy.rs
```diff
@@ -73,6 +78,33 @@ impl<'a, 'b> LogicalColumnBuilderProxy<'a, 'b> for LogicalAnyColumnBuilderProxy<
...
+}
+
+impl ColumnBuilderProxy<String> for LogicalStringColumnBuilderProxy<'_, '_> {
+    logical_generic_trait_impl!();
+
+    fn add(&mut self, input: String) -> Result<(), Error> {
+        self.commit();
+        // NOTE: we just pipe the string through as is, in particular we do not parse potential RDF terms
+        // NOTE: we store the string in the same format as it would be stored in an any column;
+        // this is important since right now we simetimes use the LogicalStringColumnBuilderProxy to directly write data that is known to only be strings into an any column and not only into string columns
```
mmarx, line 94: `sometimes`

src/main.rs
```diff
@@ -147,8 +147,12 @@ fn run(mut cli: CliApp) -> Result<(), Error> {
         for predicate in engine.program().output_predicates() {
             let mut writer = output_manager.create_file_writer(&predicate)?;
 
-            if let Some(table) = engine.table_serializer(predicate)? {
-                writer.write_trie(table)?;
+            // if let Some(table) = engine.table_serializer(predicate)? {
+            //     writer.write_trie(table)?;
+            // }
```
mmarx, line 152: Can this go?

mmarx (changes requested):

nemo/src/types.rs
```diff
@@ -126,6 +132,23 @@ impl LogicalTypeEnum {
                     }
                 }
             }
+            Self::String => {
+                match gt {
+                    Term::StringLiteral(s) => DataValueT::String(format!("\"{s}\"")),
+                    // TODO: is it correct to support language strings here omitting the language info?
+                    Term::RdfLiteral(RdfLiteral::LanguageString { value, .. }) => {
+                        DataValueT::String(format!("\"{value}\""))
+                    }
```
mmarx, line 141: A `LanguageString` is distinct from an `xsd:string` literal (which is what we treat bare string literals as), so I guess dropping the language tag here is not correct. The N-Triples reader currently stores those as `"value"@tag`.

nemo/src/io/dsv.rs
```diff
@@ -359,13 +366,18 @@ The next column is empty;;789
...
             .collect();
+        let col2: Vec<String> = col2_idx
+            .iter()
+            .copied()
+            .map(|idx| dict.get_mut().entry(idx.try_into().unwrap()).unwrap())
+            .collect();
 
-        std::iter::zip(std::iter::zip(col0, col1), col2)
-            .map(|((c0, c1), c2)| (c0, c1, *c2))
+        std::iter::zip(std::iter::zip(std::iter::zip(col0, col1), col2), col3)
```
mmarx, line 375: might be time for a `use std::iter::zip` ;)

nemo/src/api.rs
```diff
@@ -81,8 +81,12 @@ pub fn write(path: String, engine: &mut Engine, predicates: Vec<Identifier>) ->
 
     for predicate in predicates {
         let mut writer = file_manager.create_file_writer(&predicate)?;
-        let Some(serializer) = engine.table_serializer(predicate)? else { continue; };
-        writer.write_trie(serializer)?;
+        //let Some(serializer) = engine.table_serializer(predicate)? else { continue; };
+        //writer.write_trie(serializer)?;
```
mmarx, line 85: Can this go?

nemo-cli/src/main.rs
```diff
@@ -145,8 +145,12 @@ fn run(mut cli: CliApp) -> Result<(), Error> {
         for predicate in engine.program().output_predicates() {
             let mut writer = output_manager.create_file_writer(&predicate)?;
 
-            if let Some(table) = engine.table_serializer(predicate)? {
-                writer.write_trie(table)?;
+            // if let Some(table) = engine.table_serializer(predicate)? {
+            //     writer.write_trie(table)?;
+            // }
```
mmarx, line 150: Can this go?

mmarx (changes requested):

resources/testcases/ntriples/example3/atoms.csv
```diff
@@ -1 +1 @@
-http://en.wikipedia.org/wiki/Helium,"""2""^^<http://www.w3.org/2001/XMLSchema#integer>","""1.663E-4""^^<http://www.w3.org/2001/XMLSchema#double>"
\ No newline at end of file
+http://en.wikipedia.org/wiki/Helium,2,1.663E-4
```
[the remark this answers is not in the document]
mmarx: It could also be that this is not something that we should test using CSV exports, I guess we should discuss this on Wednesday.

nemo/src/types.rs
mmarx: Specifically, [RDF 1.1 concepts](https://www.w3.org/TR/rdf11-concepts/#dfn-language-tag) defines `http://www.w3.org/1999/02/22-rdf-syntax-ns#langString` as the type for language tagged strings (but explicitly does not define a datatype for this IRI).

mmarx (approved):

monsterkrampe:
Alright, I think all comments should be addressed now :)
(Hope I did not overlook something.)

Some things like output handling need more love but I think we can leave this for later. (Possibly in the scope of #245)

mmarx (approved):

PR #247, opened 2023-05-28
Change the parameter type of `ExecutionEngine` to accept `ChaseProgram`

matzemathics:
Rational: In the API we probably want to expose `ChaseProgram` rather than `Program`, but then we need to be able to construct an `ExecutionEngine` from that.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -60,7 +60,7 @@ pub fn load(file: PathBuf) -> Result<Engine, Error> {
 /// Returns an appropriate [`Error`][crate::error::Error] variant on parsing and feature check issues.
 pub fn load_string(input: String) -> Result<Engine, Error> {
     let program = all_input_consumed(RuleParser::new().parse_program())(&input)?;
-    ExecutionEngine::initialize(program)
+    ExecutionEngine::initialize(program.into())
 }
 
 /// Executes the reasoning process of the [`Engine`].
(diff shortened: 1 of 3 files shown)


mmarx:
> In the API we probably want to expose ChaseProgram rather than Program

I don't think we want to expose `ChaseProgram` _at all_.

matzemathics:
Well for the python bindings I'd at least want to have one parsed program representation, from which I can get things like what are the output predicates, the data source etc., because that should be accessible without constructing an execution engine (and also I don't want to expose a single god object).
So would it be better to expose `Program` or `ChaseProgram` or something else entirely?

matzemathics:
Btw. I mean expose not in the sense of 'reexport from api.rs', but in the sense of what our evolving entire API should look like.

mmarx:
The model we expose as part of the API should be the rule model, the chase model is for internal use only (in general, we want to do as much as possible on the rule model, since transitioning to the chase model incurs a loss of information). So, yeah, `Program` is something that can safely be exposed.

`Program` might currently lack some of the getters for, e.g., output predicates, but rather than exposing internals, we should just implement what's missing.

matzemathics:
Ok, I can see that, the only caveat is that a program now cannot be retrieved again from the execution engine, because it is transformed into an internal representation. But I guess that is fine, since Program is `Clone` anyway.

PR #249, opened 2023-05-29
create rust-toolchain.toml

matzemathics:
The `rust-toolchain.toml` sets the default toolchain to nightly, so a clean checkout of this repository now can be built without running `rustup override set ...`


rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "nightly"
+components = [ "rustfmt", "clippy", "rust-docs" ]
```
mmarx, line 3: Should probably include `miri` as well?

mmarx (approved):

PR #250, opened 2023-05-31
split repository into workspace of three crates

matzemathics:


matzemathics:
the only gnarly thing currently is error management, because errors are passed both ways between logical and physical and it is not ultimately clear to me, what belongs where.

nemo/Cargo.toml
```diff
@@ -0,0 +1,67 @@
...
+[features]
+default = ["timing"]
+# Allows building for web assembly environments
+# Enables the "js" feature of the "getrandom" crate
+# This feature cannot be used together with the "timing" feature, because the "howlong" crate does not support web assembly environments
+js = ["getrandom/js"]
+# Enables time measurements using the "howlong" crate
+# If this feature is not enabled, all time measurements will display zero instead
+timing = ["dep:howlong"]
+no-prefixed-string-dictionary = []
```
mmarx, line 23: This should be propagated to `nemo-physical`, otherwise it will just silently do nothing.

nemo/src/io/parser/types.rs
```diff
@@ -279,12 +279,22 @@ impl FromExternalError<Span<'_>, ParseError> for LocatedParseError {
 
 impl FromExternalError<Span<'_>, ParseIntError> for LocatedParseError {
     fn from_external_error(input: Span, _kind: ErrorKind, e: ParseIntError) -> Self {
-        ParseError::ExternalError(Box::new(e.into())).at(input)
+        ParseError::ExternalError(Box::new(crate::error::Error::PhysicalError(e.into()))).at(input)
     }
 }
 
 impl FromExternalError<Span<'_>, ParseFloatError> for LocatedParseError {
     fn from_external_error(input: Span, _kind: ErrorKind, e: ParseFloatError) -> Self {
```
mmarx: I'm not too happy about `ParseIntError` and `ParseFloatError` being in the physical layer, but I guess they need to be there … :/

nemo-physical/src/table_reader.rs
```diff
@@ -0,0 +1,20 @@
...
+/// # Note
+/// This is the physical interface to access all readers, instantiated on the logical layer.
+/// Therefore every reader needs to implement this trait.
+pub trait TableReader: std::fmt::Debug {
+    /// Read the table into multiple [`ColumnBuilderProxy`][crate::builder_proxy::PhysicalColumnBuilderProxy]
+    fn read_into_builder_proxies<'a: 'b, 'b>(
+        &self,
+        builder_proxies: &'b mut Vec<PhysicalBuilderProxyEnum<'a>>,
+    ) -> Result<(), Error>;
+}
```
mmarx: We might want to have a separate error type for things that could go wrong during reading (since those are shared between the logical and physical layers) – possibly also later on for writing. Even if we don't do this, I guess it might make sense to have the concrete Error type here given as an associated type, but we do require this to be object safe, don't we?

nemo/src/table_manager.rs
```diff
@@ -326,7 +326,8 @@ impl TableManager {
     pub fn table_serializer(&mut self, id: TableId) -> Result<impl TrieSerializer + '_, Error> {
         let _ = self
             .database
-            .get_trie_or_load(id, &ColumnOrder::default())?;
+            .get_trie_or_load(id, &ColumnOrder::default())
+            .map_err(|e| Error::PhysicalError(e.into()))?;
```
mmarx, line 330: Can't we have an `impl From<ReadingError> for nemo::error::Error` that does the wrapping inside `Error::PhysicalError` for us?

mmarx (changes requested):
Looks very good now, with some of the comments already outdated by the time I got through all the changes … :D

nemo-physical/src/table_reader.rs
mmarx: I like the new `ReadingError` (we might want to change the name later on, but for now this is good).

nemo/src/io/parser/types.rs
mmarx: Just to clarify, this is not a blocker, but it will make some parse errors more difficult to read. Another option might be to duplicate these two in both error types.

nemo-physical/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+"Markus Krötzsch <markus.kroetzsch@tu-dresden.de>",
+"Maximilian Marx <maximilian.marx@tu-dresden.de>"
+]	
+edition = "2021"
+
+[features]
+default = ["timing"]
+# Enables time measurements using the "howlong" crate
+# If this feature is not enabled, all time measurements will display zero instead
+timing = ["dep:howlong"]
```
mmarx, line 18: `no-prefixed-string-dictionary` still needs to be defined here, doesn't it?

nemo/Cargo.toml
mmarx: The actual dispatching happens in `nemo-physical/src/management/database.rs`, so this is where the flag ultimately needs to be set.

mmarx (changes requested):

mmarx (approved):

PR #255, opened 2023-06-05
Add nemo-wasm crate

rlwww:


mmarx (approved):

nemo-wasm/.gitignore
```diff
@@ -0,0 +1 @@
+/target
```
mmarx, line 1: Is this really still needed? Since this is now part of the workspace, artifacts should go in the top-level `target` directory, which is already ignored.

mmarx (changes requested):

mmarx (approved):

nemo-wasm/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+js-sys = "0.3.63"
+nemo = { path = "../nemo", features = [ "js" ], default-features = false }
+nemo-physical = { path = "../nemo-physical", default-features = false }
+wasm-bindgen = "0.2"
+
+[profile.release]
+lto = true
```
mmarx, line 18:
```
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
```

mmarx (changes requested):

nemo-wasm/Cargo.toml
matzemathics: Actually I am not sure, we want to enable lto globally on every release build. Maybe we should add a `release-lto` profile specifically for wasm?

mmarx: Might be a good idea. Having it below `nemo-wasm` will, however, just build the wasm without LTO, which we certainly also don't want to do.

nemo-wasm/README.md
```diff
@@ -0,0 +1,75 @@
...
+for (const predicate of program.getOutputPredicates()) {
+    const rows = new NemoResultsIterable(engine.getResult(predicate));
+
+    for (const row of rows) {
+        console.log(row);
+    }
+}
+
+// See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators
+// Iterables are not directly supported yet, see https://github.com/rustwasm/wasm-bindgen/issues/1478
```
matzemathics, line 65: I saw you do this with inline Javascript in the `nemo-web` version. Couldn't we have that here too?

nemo-wasm/Cargo.toml
mmarx: It seems there's a way to [override profiles for workspace members](https://doc.rust-lang.org/cargo/reference/profiles.html#overrides), that way we can have a `release` profile for `nemo-wasm` that has `lto`, while not needing to have it on, e.g., `nmo-cli`.

nemo-wasm/README.md
matzemathics: Ok, I guess that is also fine for now. This can always be added later.

nemo-wasm/Cargo.toml
matzemathics: Actually no:
```
error: failed to parse manifest at `/home/matze/Documents/Studium/6/knowsys/nemo/Cargo.toml`

Caused by:
  `lto` may not be specified in a `package` profile
```

mmarx: Yeah, we found that already and have agreed to take out the profile for now. Once `wasm-pack` can deal with custom profiles, we can re-enable this.

mmarx (approved):

nemo-wasm/Cargo.toml
matzemathics: In this case I would advocate for enabling `lto` globally.

Issue #257, opened 2023-06-05
Document how to use the nix flake


mmarx:
The nix flake is not currently documented. This should change.

mmarx:
Done.

PR #258, opened 2023-06-05
Add nemo-wasm to the flake

mmarx:
Expose the `nemo-wasm` bindings in the flake, and build `nemo-wasm` in the CI as well.

Depends on #255. Not covering documentation, as that is #257.


.dir-locals.el
matzemathics, line 1: Was this checked in intentionally?

mmarx: Yes.

ellmau (approved):
LGTM

PR #269, opened 2023-06-17
Support reading RDF 1.1 N-Triples files

mmarx:
Add support for N-Triples data sources, which are read as ternary predicates.


nemo/src/execution/execution_engine.rs
```diff
@@ -120,11 +120,13 @@ impl<Strategy: RuleSelectionStrategy> ExecutionEngine<Strategy> {
                         .get(predicate)
                         .cloned()
                         .unwrap_or_else(|| vec![LogicalTypeEnum::Any; arity]);
-                    let reader = DSVReader::dsv(*file.clone(), *delimiter, logical_types);
+                    let reader = DSVReader::dsv(file.to_path_buf(), *delimiter, logical_types);
 
                     TableSource::FileReader(Box::new(reader))
                 }
-                DataSource::RdfFile(_) => todo!("RDF data sources are not yet implemented"),
+                DataSource::RdfFile(file) => {
```
matzemathics, line 127: Now that we have a concrete implementation `RdfFile` seems a bit under-specified as an enum entry.

nemo/src/io.rs
```diff
@@ -44,3 +45,52 @@ impl<W: Write> RecordWriter for csv::Writer<W> {
...
+/// [`File`][`std::fs::File`].
+///
+/// ```ignore
+/// read_from_possibly_compressed_file!(file, |rdr| a)
+/// ```
+/// reads from `file`, a [`Path`][`std::path::PathBuf`]. This calls
+/// `a` with `rdr` bound to either the
+/// [`GzDecoder`][`flate2::read::GzDecoder`] or to the
+/// [`File`][`std::fs::File`].
+macro_rules! read_from_possibly_compressed_file {
```
matzemathics, line 70: Any specific reason you are not using `Box<dyn Read>` (which would make this implementable as a function) ? Anyway I think we might want to clean this up later anyway, because the decompression should probably be factored out of the individual readers.

matzemathics (approved):
Looks good to me for now, although I think we should come back to this after the 0.2 release.

nemo/src/execution/execution_engine.rs
mmarx: Why? For N-Triples, there's no degrees of freedom anyways. We definitely want to overhaul the whole `@source` syntax (to, e.g., allow for different delimiters or escape characters in DSV files), so there's certainly an opportunity to change how we handle RDF input as well, but for now, this is what rulewerk does as well.

nemo/src/io.rs
mmarx: For starters, `NTriplesReader` really needs a `BufRead`, not a `Read` (but we don't want a `BufRead` for DSV files, since the `csv` crate does it's own buffering internally). I agree, though, that this should, ideally, be handled before dispatching to the individual `TableReader`s.

PR #271, opened 2023-06-19
rewrite release action to support glibc instead of musl

matzemathics:
I also got rid of cross compilation.


mmarx:
Could you please ask before starting to workn on a PR for an issue assigned to somebody else to avoid needless duplication of work?

Also, won't this create two releases, one with each binary?

PR #274, opened 2023-06-20
Unpin nightly toolchain

mmarx:
With https://github.com/rust-lang/rust/issues/112831 fixed, we can unpin the nightly toolchain again.

Should not be merged before the first 2023-06-20 (or later) nightly build with all required components is available.


mmarx:
Since the CI checks that the toolchain is indeed unbroken, this doesn't need to be a draft.

PR #276, opened 2023-06-21
Feature/272 variable comparisons

aannleax:
Implements comparison operations between variables, i.e. you can now write `?X < ?Y` in the rule body. It should now also handle multiple comparisons correctly (before, writing `?X < 4` and `?X < 8` resulted in the first restriction being ignored).

This functionality has been implemented by extending `ColumnScanEqualValue` which has now been renamed to `ColumnScanRestrictValues` (and similar for the associated trie scan).


nemo-physical/src/columnar/operations/columnscan_restrict_values.rs
```diff
@@ -0,0 +1,355 @@
+use super::super::traits::columnscan::{ColumnScan, ColumnScanCell};
+use crate::datatypes::ColumnDataType;
+use std::{fmt::Debug, ops::Range};
+
+/// Concrete value a interval bound can take.
```
mmarx, line 5: `an interval`

nemo-physical/src/columnar/operations/columnscan_restrict_values.rs
```diff
@@ -0,0 +1,355 @@
...
+    T: Clone,
+{
+    /// Return the column index this value refers to.
+    /// Returns None if this is a constant.
+    pub fn column_index(&self) -> Option<usize> {
+        if let Self::Column(index) = self {
+            Some(*index)
+        } else {
+            None
+        }
```
mmarx, line 28: Please use an explicit `match`, so we will not silently do the wrong thing when a new variant is added to the enum.

nemo-physical/src/columnar/operations/columnscan_restrict_values.rs
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    /// Return a mutable reference to the column index this value refers to.
+    /// Returns None if this is a constant.
+    pub fn column_index_mut(&mut self) -> Option<&mut usize> {
+        if let Self::Column(index) = self {
+            Some(index)
+        } else {
+            None
+        }
```
mmarx, line 38: see above

nemo-physical/src/tabular/operations/triescan_select.rs
```diff
@@ -224,63 +227,108 @@ impl<'a> TrieScanSelectValue<'a> {
...
+                                            "Expected a column scan of type {}",
+                                            stringify!($variant)
+                                        );
+                                    };
+
+                                    scans_restriction.push(referenced_scan);
+
+                                    map_len
+                                }
+                            };
```
mmarx, line 311: Use `column_map.entry(*bound_index).or_insert_with(|| …)`

mmarx (changes requested):
Just some small comments, looks good otherwise!

nemo-physical/src/util.rs
```diff
@@ -1,6 +1,6 @@
 //! This module collects miscellaneous functionality.
 
-pub mod interval;
+// pub mod interval;
```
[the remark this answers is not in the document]
aannleax: Code there is not used anymore. I'll leave the file but remove the commented out module statement.

nemo-physical/src/util.rs
```diff
@@ -1,6 +1,5 @@
 //! This module collects miscellaneous functionality.
 
-pub mod interval;
```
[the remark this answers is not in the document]
mmarx: Still better than having dead code that is not compiled, so we will never know when it breaks ;)

mmarx (approved):

PR #277, opened 2023-06-24
Python tests

matzemathics:


nemo-python/tests/test.py
```diff
@@ -0,0 +1,79 @@
...
+        result = list(engine.result("calculated"))
+
+        expected_result = [
+            ['2', '<__Null#9223372036854775809>'],
+            ['42', '<__Null#9223372036854775810>'],
+            ['world', '<__Null#9223372036854775811>']]
+
+        self.assertEqual(result, expected_result)
+
+def stringify(value):
```
mmarx, line 31: Please make this file conform to PEP-8 (which, among other things, requires two blank lines between top-level definitions), ideally by just using https://github.com/psf/black.

matzemathics:
@mmarx I'm currently unhappy with the way I supply the testcase directory to the test script. Any ideas?

mmarx:
You could use something like 
```python
from os.path import join, basename
join(basename(__file__), '..', '..', 'resources', 'testcases')
```

nemo-python/tests/test_backbox.py
```diff
@@ -0,0 +1,68 @@
...
+                with open(results_file_name) as results_file:
+                    expected = list(csv.reader(results_file))
+
+                    for result in engine.result(relation):
+                        result = [stringify(v) for v in result]
+
+                        self.assertTrue(
+                            result in expected,
+                            f"error at {relation}",
+                        )
```
mmarx, line 48: Note that this only checks whether each result returned is contained in the expected results, not that all expected results are returned (but this can still be fine, since we're checking that in the `nemo-cli` integration tests).

mmarx (approved):

nemo-python/tests/test_backbox.py
matzemathics: Yeah, except if the language binding implementation ends up eating some results (e.g. due to value conversion or whatnot). So I will add this. Thanks for the heads-up.

mmarx (approved):

Issue #244, opened 2023-05-25
Implement Automatic Type Conversions for Rule File Terms


monsterkrampe:
There are some obvious type conversions that currently produce type errors.
For example, if you use the term `2.0` in a `Float64` column in a rule file it gives a type error, since `2.0` is parsed a `Decimal` and not `Double`. We currently only support proper `Double`s for `Float64` columns.
Before implementing this, I think there is some discussion needed on which conversions we want to allow.
Sometimes too many implicit casts may make it harder for users to spot errors.
For example it may be worthwhile to force users to put quotes around values that go into `String` columns, e.g. to write `"2"` instead of just `2`.

In any case, error messages should be improved to give suggestions like: `did you mean to write 2.0E0`?

mmarx:
My understanding would be that during parsing, we treat a numeric literal value like `2` or even `2.0` (but, notably, not `2.7` or `2.7182818284590`) as an element of some universal `Number` type that has no defined storage type, but is implicitly coercible to `Integer`, `UnsignedInteger`, `Float`, and `Double` (numeric literal values like `2.7` would only be implicitly coercible to `Float` and to `Double`, but, e.g., not to `Integer`). Then, during type checking, we would assign a “proper” numeric type to those positions (or promote them to `any`) based on the usage (we might need some default rule to resolve any ambiguities here, probably `Integer` makes the most sense as a fallback type).

mkroetzsch:
I agree with Max's proposal. The "union" type for "decimal notation number" is interesting for type inference, since it cannot be represented as an interval (float and decimal being incomparable), but this seems unavoidable (@matzemathics). I think allowing 2 to represent a float is ok (I don't see any issue with this, and it might be handy for people who process JSON where all numbers are float and you might also have a plain 2, I suppose). 

Strings in rules files should always have quotes around them, so 2 should never be a valid string.

This discussion is only about rules files. The way in which one would read, e.g., strings from CSV, can be different.

monsterkrampe:
This has been resolved by allowing implicit casts from Numeric Integer Literals and Numeric Decimal Literals to `Float64`.
The corresponding RDF Literals are not casted though.
We don't do any other automatic conversions for now.
Error messages have not been changed.

PR #281, opened 2023-06-29
allow converting decimal and integer ground terms to float64

matzemathics:


nemo/src/types.rs
```diff
@@ -143,6 +143,15 @@ impl LogicalTypeEnum {
             },
             Self::Float64 => match gt {
                 Term::NumericLiteral(NumericLiteral::Double(d)) => DataValueT::Double(d),
+                Term::NumericLiteral(NumericLiteral::Decimal(a, b)) => {
+                    let shift = f64::log10(b as f64).floor() + 1.;
+                    DataValueT::Double(Double::from_number(
+                        a as f64 + b as f64 * 10f64.powf(-shift),
+                    ))
+                }
```
mmarx, line 151: Please just use `format!("{a}.{b}").parse::<f64>()`, which will make sure that rounding happens correctly (and only once).

mmarx (changes requested):

monsterkrampe:
Is this supposed to be fixing #244 ?

nemo/src/types.rs
```diff
@@ -143,6 +143,15 @@ impl LogicalTypeEnum {
...
+                Term::NumericLiteral(NumericLiteral::Integer(a)) => {
+                    DataValueT::Double(Double::from_number(a as f64))
```
monsterkrampe, line 153: Similar handling should probably also introduced for the `RdfLiterals` that are `XSD_INTEGER` or `XSD_DECIMAL`, respectively (further below).

nemo/src/types.rs
```diff
@@ -143,11 +143,17 @@ impl LogicalTypeEnum {
...
+                }
                 Term::RdfLiteral(RdfLiteral::DatatypeValue {
                     ref value,
                     ref datatype,
                 }) => match datatype.as_str() {
-                    XSD_DOUBLE => DataValueT::Double(
+                    XSD_DOUBLE | XSD_DECIMAL | XSD_INTEGER => DataValueT::Double(
```
mmarx, line 156: Do we really want to implicitly convert integers to doubles, possibly losing precision?

matzemathics: I have no real opinion about this. Generally I am against implicit conversions, but in this case we are talking about constants somebody types into a rules file. I could add a check that the parsed integer is smaller than 2**53 or disallow integer terms in double columns entirely, but then again, if the column type was defined to be double the user willingly allows loosing precision.

matzemathics: Also there would be the possibility to allow 2 but disallow "2"^^xsd:integer. Not sure.

mmarx: > Also there would be the possibility to allow 2 but disallow "2"^^xsd:integer. Not sure.

yeah, `2` should be accepted, but if somebody bothers to write `"2"^^xsd:integer`, I would expect that they care about XSD semantics, and those clearly state that `xsd:integer` and `xsd:double` have distinct value spaces.

mmarx: Not sure how easy that would be to realise without major changes, though.

matzemathics: On second thought, it feels that if someone tries to put a "2"^^xsd:integer into a float column, they probably made a mistake. So I will remove that cast for now.

mmarx (approved):

nemo/src/types.rs
monsterkrampe: Agreed :)

monsterkrampe (approved):

PR #288, opened 2023-07-03
Add HTTP resource provider

rlwww:


nemo-physical/src/table_reader.rs
```diff
@@ -3,7 +3,13 @@
 
 use crate::{builder_proxy::PhysicalBuilderProxyEnum, error::ReadingError};
 
-/// A general interface for reading tables from files.
+/// Resource that can be reference in source declarations in Nemo programs
```
mmarx, line 6: `reference` ~> `referenced`

nemo/src/execution/execution_engine.rs
```diff
@@ -45,6 +45,8 @@ pub struct ExecutionEngine<RuleSelectionStrategy> {
 
     rule_strategy: RuleSelectionStrategy,
 
+    #[allow(dead_code)]
```
mmarx, line 48: Is this still `dead_code`? Looks like it's used below, so the attribute could go?

nemo/src/io/input_manager.rs
```diff
@@ -0,0 +1,199 @@
...
+#[derive(Debug, Clone, Copy, Default)]
+pub struct FileResourceProvider {}
+
+impl ResourceProvider for FileResourceProvider {
+    fn open_resource(&self, resource: &Resource) -> Result<Option<Box<dyn Read>>, ReadingError> {
+        // Try to parse as file IRI
+        let path = if is_iri(resource) {
+            if resource.starts_with("file:") {
+                // File URI
+                unimplemented!();
```
mmarx, line 118: This should be easy to implement, shouldn't it?

mmarx (changes requested):

nemo-python/src/lib.rs
```diff
@@ -104,7 +104,8 @@ struct NemoEngine(nemo::execution::DefaultExecutionEngine);
 impl NemoEngine {
     #[new]
     fn py_new(program: NemoProgram) -> PyResult<Self> {
-        let engine = ExecutionEngine::initialize(program.0).py_res()?;
+        let engine =
+            ExecutionEngine::initialize(program.0, ResourceProviders::from(vec![])).py_res()?;
```
matzemathics, line 108: This should probably be `ResourceProviders::default`, or be controlled by the caller (maybe an optional argument). Otherwise the test cases (integration tests) will fail.

matzemathics:
By the way if you want to run the python tests via nix (without setting up a python venv for development) you can use
```
nix run .#python -m unittest discover -s nemo-python/tests -v
```
(which should do the same thing as the ci)

matzemathics (approved):
Overall I like the approach. I think the possibility to control the resource providers via python is probably out of scope for this PR, so I propose you just use `ResourceProviders::default` for now.

mmarx (approved):

PR #289, opened 2023-07-03
fix CI errors

matzemathics:


mmarx (approved):

matzemathics:
I must say I find `#[warn(clippy::single_range_in_vec_init)]` pretty surprising considering rust has a very strict type system...

nemo-physical/src/lib.rs
```diff
@@ -18,6 +18,7 @@
     clippy::cast_possible_truncation,
     clippy::cast_possible_wrap
 )]
+#![allow(clippy::single_range_in_vec_init)]
```
mmarx, line 21: Please only allow this on the lines where this currently happens.

mmarx (changes requested):

mmarx:
> […] considering rust has a very strict type system...

Does the lint also trigger if there is an explicit `Vec<Range>` type annotation?

nemo-physical/src/lib.rs
matzemathics: Well as discussed above, I do not find this lint particularly useful as it is meant more as a UX improvement, to help if beginners that do not understand the error message of the type system. It does not really catch any new errors.

Additionally we happen to use `Vec<Range>` pretty often, so the code could become cluttered.
I'd say it's safe to disable this lint.

matzemathics: Basically what I'm saying is this Lint will only ever give us false-positives.

matzemathics:
> Does the lint also trigger if there is an explicit `Vec<Range>` type annotation?

Yep.
```
warning: a `Vec` of `Range` that is only one element
   --> nemo/src/table_manager.rs:663:50
    |
663 |         let expected_ranges: Vec<Range<usize>> = vec![1..5];
    |                                                  ^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#single_range_in_vec_init
    = note: `#[warn(clippy::single_range_in_vec_init)]` on by default
help: if you wanted a `Vec` that contains the entire range, try
    |
663 |         let expected_ranges: Vec<Range<usize>> = (1..5).collect::<std::vec::Vec<usize>>();
    |                                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
help: if you wanted a `Vec` of len 5, try
    |
663 |         let expected_ranges: Vec<Range<usize>> = vec![1; 5];
```

nemo-physical/src/lib.rs
mmarx: Counterpoint: we have a lot of `Vec<Range<…>>` in our code, yet we only have three occurrences of this pattern, all of them in tests, so this clearly doesn't lead to _many_ false-positives. I also don't see why this wouldn't be a relevant warning for non-`Range`-typed sequences.

To avoid clutter, though, instead of allowing this lint, just be explicit about creating single-element Vecs:
`vec![3..9; 1]` does _not_ trigger the lint _and_ communicates intent.

matzemathics: Agreed, `vec![3..9; 1]` is a valid solution I did not think of (although I'd be worried, that cargo fmt someday decides to "clean up" the unnecessary length).

I also don't find the lint entirely useless, but it seems to strong currently. I think I'll file an issue with clippy and in the mean time go with the solution you proposed.

nemo-physical/src/columnar/operations/columnscan_reorder.rs
```diff
@@ -26,7 +26,7 @@ where
 {
     /// Construct a new [`ColumnScanReorder`].
     pub fn new(column: &'a ColumnEnum<T>) -> Self {
-        ColumnScanReorder::narrowed(column, vec![0..column.len()])
+        ColumnScanReorder::narrowed(column, vec![0..column.len(); 1])
```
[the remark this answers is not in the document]
mmarx:
```
❯ cargo clippy --version
clippy 0.1.72 (839e9a6 2023-07-02)
```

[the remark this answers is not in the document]
matzemathics: ok, that is weird, i'm using the exact same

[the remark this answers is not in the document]
matzemathics:
```
cargo clippy --all --all-targets
warning: a `Vec` of `Range` that is only one element
  --> nemo-physical/src/columnar/operations/columnscan_reorder.rs:29:45
   |
29 |         ColumnScanReorder::narrowed(column, vec![0..column.len()])
   |                                             ^^^^^^^^^^^^^^^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#single_range_in_vec_init
   = note: `#[warn(clippy::single_range_in_vec_init)]` on by default
help: if you wanted a `Vec` that contains the entire range, try
   |
29 |         ColumnScanReorder::narrowed(column, (0..column.len()).collect::<std::vec::Vec<usize>>())
   |                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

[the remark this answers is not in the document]
mmarx: Oh, nevermind, not sure how I came to the conclusion that it doesn't trigger, it does indeed.

PR #290, opened 2023-07-04
Use Rio for parsing N-Triples, Turtle, RDF/XML

mmarx:
This uses the N-Triples parser from https://github.com/oxigraph/rio for parsing N-Triples, closing the difference between CSV parsing (with `any` columns) and N-Triples parsing.

Once #288 lands, we can further extend this to reading, e.g., Turtle as well.


matzemathics (approved):
Ok, looks good.

PR #295, opened 2023-07-05
Allow column-wise iteration of full tables

### LINKED ISSUE #294: Allow column-wise Iteration over full tables in physical database
As of now, the physical database only offers `Trie`s to the outside on which one can call `.records` to get an interator over the full rows of the trie. This is not ideal since all values are necessarily wrapped into `DataValueT` (if not directly transferred to `String`).

The database should also offer a method that provides a list of `Iterator`s over the columns of a full table; that is, values in the trie columns are repeated if necessary.

This can be thought of as the inverse direction of the `BuilderProxy`s.

monsterkrampe:
closes #294

### DIFF
--- a/nemo-physical/src/datatypes/data_value.rs
+++ b/nemo-physical/src/datatypes/data_value.rs
@@ -63,3 +63,20 @@ impl std::fmt::Display for DataValueT {
         }
     }
 }
+
+/// Iterator over one kind of possible data values
+#[allow(missing_debug_implementations)]
+pub enum DataValueIteratorT<'a> {
+    /// String Variant
+    String(Box<dyn Iterator<Item = String> + 'a>),
+    /// U32 Variant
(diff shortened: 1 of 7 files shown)


nemo-physical/src/datatypes/data_value.rs
```diff
@@ -63,3 +63,20 @@ impl std::fmt::Display for DataValueT {
         }
     }
 }
+
+/// Iterator over one kind of possible data values
+#[allow(missing_debug_implementations)]
```
mmarx, line 68: I guess a small handwritten `Debug` implementation couldn't hurt (it would just state which variant it's an iterator over), but for now this is also okay.

mmarx (approved):

matzemathics (approved):
lgtm

PR #297, opened 2023-07-05
Implement abstraction for mapping from logical to physical layer and vice-versa

### LINKED ISSUE #245: Implement Mapping from Physical Representation of Any to RDF Terms
The RDF terms in the logical `Any` columns are stored in a string representation in the physical layer. Currently, our output just writes this representation to CSV since it happens to be exactly what we want to have in the CSV.
In general, we want the possibility to get the original RDF term back from the physical string representation.
To be more precise, we want something like the inverse of `LogicalTypeEnum::ground_term_to_data_value_t`.
To be super clean, we would want to use this method also when writing CSV files but we have to be careful to not create wrapper objects for every value since CSV writing is very much performance critical. Maybe an idea similar to the `BuilderProxy` solution could be helpful.

### LINKED ISSUE #280: Use Logical Types and Source Types to decide which Builder Proxy to use for CSV and other Readers
As of now, the CSV reader only accepts a list of logical types and decides which builder proxies to use only based on that.
Additionally there is some hack write know that still uses the `String` Builder Proxy if the logical type is `Any` and the corresponding source type is `String`.
To make this more flexible, one could maybe pass Proxy Builders to the reader directly instead of only the logical types.

monsterkrampe:
closes #245 

I noticed that it makes sense to adjust the physical representation of the any type here so I will also touch the input handling:
closes #280

### DIFF
--- a/nemo-physical/src/dictionary/value_serializer.rs
+++ b/nemo-physical/src/dictionary/value_serializer.rs
@@ -1,7 +1,7 @@
 use std::{fmt::Display, ops::Deref};
 
 use crate::{
-    datatypes::{DataTypeName, DataValueT, StorageValueT},
+    datatypes::{data_value::PhysicalString, DataTypeName, DataValueT, StorageValueT},
     management::database::Dict,
     tabular::traits::table_schema::TableSchema,
 };
@@ -13,7 +13,7 @@ use super::Dictionary;
 pub const NULL_PREFIX: &str = "NULL:";
 
 /// Load constant from dictionary with fallback if it is not found
-pub fn serialize_constant_with_dict<C, D>(constant: C, dict: D) -> String
+pub fn serialize_constant_with_dict<C, D>(constant: C, dict: D) -> PhysicalString
 where
     usize: TryFrom<C>,
     C: Copy + Display,
@@ -23,6 +23,7 @@ where
         .ok()
         .and_then(|constant| dict.entry(constant))
         .unwrap_or_else(|| format!("{NULL_PREFIX}{constant}"))
+        .into()
 }
 
 /// Helper trait for mapping [`StorageValueT`] back into some (higher level) value space
--- a/nemo-physical/src/dictionary/value_serializer.rs
+++ b/nemo-physical/src/dictionary/value_serializer.rs
@@ -9,6 +9,9 @@ use std::convert::TryFrom;
 
 use super::Dictionary;
 
+/// Prefix for physical null representation
+pub const NULL_PREFIX: &str = "NULL:";
+
 /// Load constant from dictionary with fallback if it is not found
 pub fn serialize_constant_with_dict<C, D>(constant: C, dict: D) -> String
 where
@@ -19,7 +22,7 @@ where
     usize::try_from(constant)
         .ok()
         .and_then(|constant| dict.entry(constant))
-        .unwrap_or_else(|| format!("<__Null#{constant}>"))
+        .unwrap_or_else(|| format!("{NULL_PREFIX}{constant}"))
 }
 
 /// Helper trait for mapping [`StorageValueT`] back into some (higher level) value space
--- a/nemo-python/src/lib.rs
+++ b/nemo-python/src/lib.rs
@@ -1,6 +1,7 @@
 use std::{collections::HashSet, fs::read_to_string};
 
 use nemo::{
+    datatypes::Double,
     execution::ExecutionEngine,
     io::{resource_providers::ResourceProviders, OutputFileManager, RecordWriter},
     model::{types::primitive_logical_value::PrimitiveLogicalValueT, NumericLiteral, Term},
@@ -97,9 +98,11 @@ impl NemoResults {
                         Term::StringLiteral(s) => s.into_py(slf.py()),
                         Term::RdfLiteral(lit) => lit.to_string().into_py(slf.py()),
                     },
-                    PrimitiveLogicalValueT::String(s) => s.into_py(slf.py()),
-                    PrimitiveLogicalValueT::Integer(i) => i.into_py(slf.py()),
-                    PrimitiveLogicalValueT::Float64(d) => f64::from(d).into_py(slf.py()),
+                    PrimitiveLogicalValueT::String(s) => String::from(s).into_py(slf.py()),
+                    PrimitiveLogicalValueT::Integer(i) => i64::from(i).into_py(slf.py()),
+                    PrimitiveLogicalValueT::Float64(d) => {
+                        f64::from(Double::from(d)).into_py(slf.py())
+                    }
                 })
                 .collect(),
         )
--- a/nemo-python/src/lib.rs
+++ b/nemo-python/src/lib.rs
@@ -1,9 +1,9 @@
 use std::{collections::HashSet, fs::read_to_string};
 
 use nemo::{
-    datatypes::{DataValueT, Double, Float},
     execution::ExecutionEngine,
     io::{resource_providers::ResourceProviders, OutputFileManager, RecordWriter},
+    model::types::primitive_logical_value::PrimitiveLogicalValueT,
 };
 
 use pyo3::{create_exception, prelude::*};
@@ -71,7 +71,7 @@ impl NemoOutputManager {
 }
 
 #[pyclass]
-struct NemoResults(Box<dyn Iterator<Item = Vec<DataValueT>> + Send>);
+struct NemoResults(Box<dyn Iterator<Item = Vec<PrimitiveLogicalValueT>> + Send>);
 
 #[pymethods]
 impl NemoResults {
@@ -85,12 +85,11 @@ impl NemoResults {
         Some(
             next.into_iter()
                 .map(|v| match v {
-                    DataValueT::String(s) => s.into_py(slf.py()),
-                    DataValueT::U32(n) => n.into_py(slf.py()),
-                    DataValueT::U64(n) => n.into_py(slf.py()),
-                    DataValueT::I64(n) => n.into_py(slf.py()),
-                    DataValueT::Float(n) => (<Float as Into<f32>>::into(n)).into_py(slf.py()),
-                    DataValueT::Double(n) => (<Double as Into<f64>>::into(n)).into_py(slf.py()),
+                    // TODO: probably this should not just convert the rdf literal to a string...
+                    PrimitiveLogicalValueT::Any(rdf) => rdf.to_string().into_py(slf.py()),
+                    PrimitiveLogicalValueT::String(s) => s.into_py(slf.py()),
+                    PrimitiveLogicalValueT::Integer(i) => i.into_py(slf.py()),
+                    PrimitiveLogicalValueT::Float64(d) => f64::from(d).into_py(slf.py()),
                 })
                 .collect(),
         )
--- a/nemo-python/src/lib.rs
+++ b/nemo-python/src/lib.rs
@@ -3,7 +3,7 @@ use std::{collections::HashSet, fs::read_to_string};
 use nemo::{
     execution::ExecutionEngine,
     io::{resource_providers::ResourceProviders, OutputFileManager, RecordWriter},
-    model::types::primitive_logical_value::PrimitiveLogicalValueT,
+    model::{types::primitive_logical_value::PrimitiveLogicalValueT, NumericLiteral, Term},
 };
 
 use pyo3::{create_exception, prelude::*};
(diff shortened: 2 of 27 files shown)


nemo/src/builder_proxy.rs
```diff
@@ -11,286 +18,491 @@ use nemo_physical::{
...
-    fn new(physical_builder_proxy: &'b mut PhysicalBuilderProxyEnum<'a>) -> Self {
-        match physical_builder_proxy {
-            PhysicalBuilderProxyEnum::I64(physical) => Self { physical },
-            _ => unreachable!("If the database representation of the logical types is correct, we never reach this branch.")
-        }
+impl ColumnBuilderProxy<Double> for LogicalIntegerColumnBuilderProxy<'_> {
+    logical_generic_trait_impl!();
+
+    fn add(&mut self, input: Double) -> Result<(), ReadingError> {
+        <Self as ColumnBuilderProxy<String>>::commit(self);
```
mmarx, line 224: `<String>` ~> `<Double>`

nemo/src/builder_proxy.rs
```diff
@@ -11,286 +18,491 @@ use nemo_physical::{
...
+
+impl<T> ColumnBuilderProxy<String> for GenericLogicalParser<Term, T>
+where
+    T: ColumnBuilderProxy<Term>,
+{
+    logical_generic_trait_impl!();
+
+    fn add(&mut self, input: String) -> Result<(), ReadingError> {
+        self.inner.add(parse_rdf_term_from_string(input))
+    }
```
mmarx, line 393: Should this not commit?

nemo/src/builder_proxy.rs
```diff
@@ -11,286 +18,491 @@ use nemo_physical::{
...
+        self.inner.add(any_double_to_physical_double(input)?)
+    }
+}
+
+impl ColumnBuilderProxy<Double> for LogicalFloat64ColumnBuilderProxy<'_> {
+    logical_generic_trait_impl!();
+
+    fn add(&mut self, input: Double) -> Result<(), ReadingError> {
+        self.inner.add(input)
+    }
```
mmarx, line 270: Should this not commit?

nemo/src/io/input_manager.rs
```diff
@@ -49,6 +51,7 @@ impl InputManager {
                     self.resource_providers.clone(),
                     resource.clone(),
                     base.clone(),
+                    logical_types,
```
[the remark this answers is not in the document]
matzemathics: (i. e. the intersection)

mmarx (changes requested):
Actually, upon experimenting, it seems that the `GenericLogicalParser` impls should always `commit` before `adding`, the `Logical*ColumnBuilderProxy` impls don't, since the physical impls will always `commit` before `adding`.

mmarx:
Performance-wise, this seems to be no worse than the implementation in 0.3.0, at least on my machine.

nemo/src/builder_proxy.rs
```diff
@@ -11,286 +18,491 @@ use nemo_physical::{
...
 }
 
 impl ColumnBuilderProxy<String> for LogicalFloat64ColumnBuilderProxy<'_> {
     logical_generic_trait_impl!();
 
     fn add(&mut self, input: String) -> Result<(), ReadingError> {
-        self.commit();
-        self.physical.add(Double::new(input.parse::<f64>()?)?)
+        <Self as ColumnBuilderProxy<String>>::commit(self);
```
mmarx, line 277: This doesn't need to commit either.

nemo/src/builder_proxy.rs
```diff
@@ -11,286 +18,491 @@ use nemo_physical::{
...
+    {
+        GenericLogicalParser::new(self)
+    }
+}
+
+impl ColumnBuilderProxy<Term> for LogicalIntegerColumnBuilderProxy<'_> {
+    logical_generic_trait_impl!();
+
+    fn add(&mut self, input: Term) -> Result<(), ReadingError> {
+        <Self as ColumnBuilderProxy<Term>>::commit(self);
```
mmarx, line 198: This doesn't need to commit.

nemo/src/builder_proxy.rs
```diff
@@ -11,286 +18,491 @@ use nemo_physical::{
...
 
-/// [`LogicalColumnBuilderProxy`] to add String
-#[derive(Debug)]
-pub struct LogicalStringColumnBuilderProxy<'a: 'b, 'b> {
-    physical: &'b mut PhysicalStringColumnBuilderProxy<'a>,
+impl ColumnBuilderProxy<Term> for LogicalStringColumnBuilderProxy<'_, '_> {
+    logical_generic_trait_impl!();
+
+    fn add(&mut self, input: Term) -> Result<(), ReadingError> {
+        <Self as ColumnBuilderProxy<Term>>::commit(self);
```
mmarx, line 140: This doesn't need to commit.

nemo/src/builder_proxy.rs
```diff
@@ -11,286 +18,491 @@ use nemo_physical::{
...
-            // it is, pass as-is
-            return trimmed.to_string();
-        }
+impl ColumnBuilderProxy<Term> for LogicalAnyColumnBuilderProxy<'_, '_> {
+    logical_generic_trait_impl!();
 
-        // otherwise it needs to be quoted
-        format!(r#""{trimmed}""#).to_string()
+    fn add(&mut self, input: Term) -> Result<(), ReadingError> {
+        <Self as ColumnBuilderProxy<Term>>::commit(self);
```
mmarx, line 84: This doesn't need to commit.

matzemathics (changes requested):
I have a few comments, overall looks really solid. I haven't seen any new tests for the mapping from the physical layer back to the logical layer. If they don't land as part of this PR I suggest creating a new issue for that.

monsterkrampe:
Thanks for the thorough reviews and testing performance!
There should definitely be tests, I think this is well in scope of the PR and I should take the time to write them.
As of now I mostly used the CSV and RDF Triples tests to also transform the input back with the `OutputMapper`s and match the (largely unchanged) expected values but this is not optimal.

nemo/src/model/rule_model/term.rs
```diff
@@ -118,10 +115,24 @@ pub enum Term {
 impl std::fmt::Display for Term {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match &self {
-            Term::Constant(term) => write!(f, "{term}"),
+            Term::Constant(Identifier(s)) => {
+                // Nulls start with __ and shall be wrapped in angle brackets
+                if s.starts_with("__") {
+                    write!(f, "<{s}>")
+                }
```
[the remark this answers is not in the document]
monsterkrampe: This has the same output as the else. I could add this as a negated condition to the `else if` as well and get rid of the extra branch.

I think on logical level the nulls start with `__Null#` (as before), the `NULL_PREFIX` only concerns the physical layer.

nemo/src/io/formats/rdf_triples.rs
```diff
@@ -7,68 +7,59 @@ use nemo_physical::{
...
-        } else {
-            //
-            self.0
-        }
+pub(crate) fn rio_term_to_term(input: rio_api::model::Term) -> Result<Term, ReadingError> {
+    match input {
+        rio_api::model::Term::NamedNode(nn) => Ok(rio_named_node_to_term(nn)),
+        rio_api::model::Term::BlankNode(bn) => Ok(rio_blank_node_to_term(bn)),
+        rio_api::model::Term::Literal(lit) => Ok(rio_literal_to_term(lit)),
+        rio_api::model::Term::Triple(_t) => Err(ReadingError::RdfStarUnsupported),
```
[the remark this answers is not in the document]
monsterkrampe: The only reason why I decided against implemting `(Try)From` here was that if we at some point decide to put `io` into its own crate, then we are not allowed to implement foreign traits for foreign types. I do not know how likely this is.
I'm happy to change this (for now) :)

nemo/src/builder_proxy.rs
monsterkrampe: (Unless I missed something), I'm always committing manually if an error might occur before the inner `add` is called.
I think this is necessary to prevent rollbacks of input rows that have been read completely.
We could unify this to commit manually everywhere; wouldn't hurt.

nemo/src/builder_proxy.rs
monsterkrampe: Good catch!

nemo-python/src/lib.rs
```diff
@@ -85,12 +85,21 @@ impl NemoResults {
...
-                    DataValueT::Float(n) => (<Float as Into<f32>>::into(n)).into_py(slf.py()),
-                    DataValueT::Double(n) => (<Double as Into<f64>>::into(n)).into_py(slf.py()),
+                    PrimitiveLogicalValueT::Any(rdf) => match rdf {
+                        Term::Variable(_) => panic!("Variables should not occur as results!"),
+                        Term::Constant(c) => c.to_string().into_py(slf.py()),
+                        Term::NumericLiteral(NumericLiteral::Integer(i)) => i.into_py(slf.py()),
+                        Term::NumericLiteral(NumericLiteral::Double(d)) => {
+                            f64::from(d).into_py(slf.py())
+                        }
+                        // currently we pack decimals into strings, maybe this should change
```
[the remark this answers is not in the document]
monsterkrampe: Yes! I will create a new issue for the python (and wasm) bindings to make better use of the exposed Rdf Terms. (I'm not seeing this so much in the scope of this PR and just implemented something that works without putting too much love into it.)

nemo/src/model/rule_model/term.rs
```diff
@@ -118,10 +115,24 @@ pub enum Term {
 impl std::fmt::Display for Term {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match &self {
-            Term::Constant(term) => write!(f, "{term}"),
+            Term::Constant(Identifier(s)) => {
+                // Nulls start with __ and shall be wrapped in angle brackets
```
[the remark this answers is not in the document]
monsterkrampe: (As noted in another comment above:) The `NULL_PREFIX` only concerns the physical layer. At this point, we consider the logical representation which should still start with `__Null#` as before if I'm not mistaken.

nemo-physical/src/dictionary/value_serializer.rs
```diff
@@ -9,6 +9,9 @@ use std::convert::TryFrom;
 
 use super::Dictionary;
 
+/// Prefix for physical null representation
+pub const NULL_PREFIX: &str = "NULL:";
```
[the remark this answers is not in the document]
monsterkrampe: No, this is intentional. All the physical prefixes are of the form `TYPE:`.

nemo/src/io/formats/rdf_triples.rs
```diff
@@ -109,16 +103,26 @@ impl RDFTriplesReader {
...
+            .zip(self.logical_types.clone())
+            .map(|(bp, lt)| {
+                let boxed: Box<dyn ColumnBuilderProxy<Term>> =
+                    match lt.wrap_physical_column_builder(bp) {
+                        LogicalColumnBuilderProxyT::Any(lcbp) => Box::new(lcbp),
+                        LogicalColumnBuilderProxyT::String(lcbp) => Box::new(lcbp),
+                        LogicalColumnBuilderProxyT::Integer(lcbp) => Box::new(lcbp),
+                        LogicalColumnBuilderProxyT::Float64(lcbp) => Box::new(lcbp),
+                    };
+                boxed
```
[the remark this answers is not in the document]
monsterkrampe: I was thinking about it and since you bring it up, I think I will go for it :)

nemo/src/io/input_manager.rs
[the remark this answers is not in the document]
monsterkrampe: The RDF sources should not allow to specify input types (the input types are always `any`).
I think as of now one can technically still specify them but they will silently be ignored.
I think we had discussions offline that each source type should get its own parameters in the source specification and then it should not be possible anymore to specify source types for RDF sources.

nemo/src/io/formats/dsv.rs
```diff
@@ -128,13 +153,15 @@ impl DSVReader {
         resource: Resource,
         delimiter: u8,
         logical_types: Vec<PrimitiveType>,
+        input_type_constraint: TupleConstraint,
```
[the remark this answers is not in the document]
monsterkrampe: I will give it some thought and change this :)

nemo/src/builder_proxy.rs
mmarx: Ah, you're right, I had not though about errors.

monsterkrampe:
(Additional Tests are still missing.)

nemo/src/builder_proxy.rs
```diff
@@ -11,286 +18,549 @@ use nemo_physical::{
...
 
 impl ColumnBuilderProxy<String> for LogicalStringColumnBuilderProxy<'_, '_> {
     logical_generic_trait_impl!();
 
     fn add(&mut self, input: String) -> Result<(), ReadingError> {
-        self.commit();
+        <Self as ColumnBuilderProxy<String>>::commit(self);
         // NOTE: we just pipe the string through as is, in particular we do not parse potential RDF terms
         // NOTE: we store the string in the same format as it would be stored in an any column;
         // this is important since right now we sometimes use the LogicalStringColumnBuilderProxy to directly write data that is known to only be strings into an any column and not only into string columns
```
mmarx, line 203: This is no longer true now, is it?

nemo/src/builder_proxy.rs
```diff
@@ -11,286 +18,723 @@ use nemo_physical::{
...
-    /// # Panics
-    /// If the logical and the nested physical type are not compatible, an `unreachable` panic will be thrown.
-    fn new(physical_builder_proxy: &'b mut PhysicalBuilderProxyEnum<'a>) -> Self
-    where
-        Self: Sized;
-}
+use super::model::types::primitive_logical_value::{
+    any_double_to_physical_double, any_integer_to_physical_integer, any_string_to_physical_string,
+    any_term_to_physical_string, logical_double_to_physical_string,
+    logical_integer_to_physical_string, logical_string_to_physical_string,
```
mmarx, line 29: I wonder whether a newtype `PhysicalString(…)` and `From`/`TryFrom` impls would be easier to read?

monsterkrampe:
Tests are there now :)

nemo/src/builder_proxy.rs
monsterkrampe: I think the first two lines are still kind of true but the comment does not really fit here anymore so I will remove it.

nemo/src/builder_proxy.rs
monsterkrampe: Done!
This got a little out of hand though introducing wrapper structs for many logical types as well.
(Note that I cannot implement e.g. `From<i64> for PhysicalString` since both `i64` and `PhysicalString` are not in the logical crate.)

mmarx (approved):
Well worth the effort, looks good!

PR #303, opened 2023-07-06
Relax ordering requirements in rules programs

### LINKED ISSUE #284: Relax ordering requirements for declarations and statements in rule files; improve error messages
Example:
```
@source a[3]: load-csv("something.csv").
@declare a(integer, string, any).
```
gives error
```
Parse error on line 2, column 1: Syntax error: End of file
```

Expected behaviour:
Either accept the above, or something like
```
Parse error on line 2, column 1: Syntax error: @declare used after @source
```

mmarx:
Fixes #284.

### DIFF
--- a/nemo/src/io/parser.rs
+++ b/nemo/src/io/parser.rs
@@ -1659,4 +1659,42 @@ mod test {
             )
         );
     }
+
+    #[test]
+    fn program_statement_order() {
+        assert_matches!(
+            parse_program(
+                r#"@output s .
+                   s(?s, ?p, ?o) :- t(?s, ?p, ?o) .
+                   @source t[3]: load-rdf("triples.nt") .
+                  "#
+            ),
+            Ok(_)
+        );
+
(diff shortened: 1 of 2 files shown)


matzemathics (approved):
Ok looks good. Maybe an end-to-end test would be nice?

nemo/src/io/parser.rs
```diff
@@ -1611,4 +1659,42 @@ mod test {
...
+    #[test]
+    fn program_statement_order() {
+        assert_matches!(
+            parse_program(
+                r#"@output s .
+                   s(?s, ?p, ?o) :- t(?s, ?p, ?o) .
+                   @source t[3]: load-rdf("triples.nt") .
+                  "#
+            ),
+            Ok(_)
```
mmarx, line 1672: We have this, not sure if an end-to-end test would provide any additional value.

matzemathics: yes very much much optional

PR #304, opened 2023-07-07
Quick fix physical type inference

aannleax:


monsterkrampe (approved):
I can't say that I fully understand what the problem is but I think this can be merged.
Please open an issue at least for the remaining TODO though. (If you havn't done this already.)

PR #306, opened 2023-07-07
Speed up loading of `any` Columns

mmarx:
For RDF-based formats, we are currently performing a redundant revalidation that the parsed terms are actually valid encodings of RDF literals. However, we already know that they are (since they were parsed successfully). This skips this validation, leading to significantly faster loading times.

For DSV files, we do not know already if the contents of a column are valid encodings of RDF files, so we optimistically try to parse those as objects in a single Turtle triple – if it succeeds, it is a valid term, and we can proceed as above, otherwise, we already know it's not valid, so it can only be an un-bracketed IRI or it needs quoting.

### DIFF
--- a/nemo/src/builder_proxy.rs
+++ b/nemo/src/builder_proxy.rs
@@ -1,6 +1,8 @@
 //! The logical builder proxy concept allows to transform a given String, representing some data in a logical datatype
 //! into some value, which can be given to the physical layer to store the data accordingly to its type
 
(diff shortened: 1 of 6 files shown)


matzemathics (approved):
Looks fine. I wonder whether it might be beneficial at some point to switch from `String` to `Arc<str>`, for dictionary entries, because those have a faster clone and we don't need mutability (I think)

PR #313, opened 2023-07-13
Roll back when failing term conversions in RDF parser

mmarx:
The RDF parsing code does not correctly handle failing term conversions when adding to the columns, leading to states where, e.g., a malformed predicate adds values to the subject and predicate columns, but not to the object column. This fixes the issue.


monsterkrampe (approved):
Totally missed that, thanks!

PR #314, opened 2023-07-13
Correctly handle pointless decimal literals

mmarx:
Decimal literals can omit the decimal point if the fractional part is zero. This was not handled before, leading to literals wrongly ignored as malformed. Since such pointless decimal literals are also precisely [valid integer literals](https://www.w3.org/TR/xmlschema11-2/#integer), this also adds the necessary conversions.


monsterkrampe (approved):
Wasn't aware of this; thank you :)

PR #317, opened 2023-07-14
Expose CI checks as flake checks

mmarx:
This exposes the various checks run in CI as flake checks, allowing `nix flake check` to test them.

The `miri` check is a bit fragile, as `miri` is very picky about the versions of its dependencies, yet doesn't support locking them in a useful way, so we patch `Cargo.lock` for that.


mmarx:
On second thought, maybe it's better not to expose `miri` as a flake check and have that still run separately. That would save a lot of trouble when updating `Cargo.lock`.

matzemathics (commented):
I'd also say that `miri` is the least useful to run as part of `flake check`, b/c the unsafe code is hardly ever touched and it's still run in CI anyway. So I suggest removing it and the patch.

matzemathics (approved):

Issue #311, opened 2023-07-13
Implement meaningful Benchmarks that can be used as performance regression tests


monsterkrampe:
We technically have benchmarks right now but those are not meaningful.
We should write benchmarks that reflect common use cases to be able to use them as regression tests before merging PRs.
It can still be decided if this should be integrated into CI.

PR #319, opened 2023-07-17
Add simple benchmarks for in-/output handling

monsterkrampe:
related to #311 

A quick idea for benchmarks for the logical input-/output mappings.


nemo-benches/benches/output.rs
```diff
@@ -0,0 +1,138 @@
...
+                    .serialize_output(DataValueIteratorT::String(Box::new(output.into_iter())))
+                {
+                }
+            },
+            BatchSize::SmallInput,
+        )
+    });
+
+    serialized_output.finish();
+}
```
monsterkrampe, line 135: Interestingly, out of these two the latter is noticeably faster. I did not investigate this too much but I would have expected them to have more or less the same performance.

mmarx (approved):

PR #320, opened 2023-07-17
Dynamic source loading

matzemathics:
Only open the source files when they are actually needed.

### DIFF
--- a/nemo-physical/src/table_reader.rs
+++ b/nemo-physical/src/table_reader.rs
@@ -4,7 +4,7 @@
 use crate::{builder_proxy::PhysicalBuilderProxyEnum, error::ReadingError};
 
 /// Resource that can be reference in source declarations in Nemo programs
-/// Resources are resolved using `nemo::io::input_manager::ResourceProviders`
+/// Resources are resolved using `nemo::io::resource_providers::ResourceProviders`
 ///
 /// Resources currently can be either an IRI or a (possibly relative) file path.
 pub type Resource = String;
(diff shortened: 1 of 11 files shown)


monsterkrampe:
I appreciate the refactoring!
From skimming through, I fail to see though how this PR changes the behavior into:
>Only open the source files when they are actually needed.


Wasn't this the case before already? (Maybe I'm really just not seeing it.)

matzemathics:
> I appreciate the refactoring! From skimming through, I fail to see though how this PR changes the behavior into:
> 
> > Only open the source files when they are actually needed.
> 
> Wasn't this the case before already? (Maybe I'm really just not seeing it.)

Yes, you are right (I just double checked). So the only big change is that the `TableReader` trait now consumes the reader when reading. I guess that also makes the `NativeDataSourceResolver` sort of obsolete.

nemo-physical/src/table_reader.rs
```diff
@@ -4,7 +4,7 @@
 use crate::{builder_proxy::PhysicalBuilderProxyEnum, error::ReadingError};
 
 /// Resource that can be reference in source declarations in Nemo programs
-/// Resources are resolved using `nemo::io::input_manager::ResourceProviders`
+/// Resources are resolved using `nemo::io::resource_providers::ResourceProviders`
```
mmarx, line 7: I guess this should be a link (if it were, we would have gotten a warning already for the target not existing)

nemo/src/model/rule_model/datasource.rs
```diff
@@ -215,7 +215,7 @@ impl DataSource for SparqlQuery {
 
 /// An external data source.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub enum DataSourceT {
+pub enum NativeDataSource {
```
mmarx, line 218: What would the non-native data sources be that we want to support?

matzemathics: I'd like to build some integration to input python tuples into the database, without going through the resource provider trait (which only works for serialized data).

monsterkrampe (approved):

PR #331, opened 2023-07-27
Add fallback for conversion of large numeric datavalue literals

### LINKED ISSUE #321: Large integer/decimal values are not accepted in parsing
(Seen on commit 4805f19)

When loading RDF data from Wikidata we encounter some values that fail to convert to type any

The following is a log entry of such case
```
INFO  nemo::io::formats::rdf_triples] Ignoring malformed triple: Failed to convert value "+9950000000000000000"^^<http://www.w3.org/2001/XMLSchema#decimal> to type any.
```

Attached you can find some RDF triples, a rule file that would load the data and extract the value in the object position, and the log produced when running nemo

[triples.nt.gz](https://github.com/knowsys/nemo/files/12094760/triples.nt.gz)
[rules.rls.gz](https://github.com/knowsys/nemo/files/12094764/rules.rls.gz)
[log.log.gz](https://github.com/knowsys/nemo/files/12094782/log.log.gz)

monsterkrampe:
fixes #321

### DIFF
--- a/nemo/src/model/types/primitive_logical_value.rs
+++ b/nemo/src/model/types/primitive_logical_value.rs
@@ -9,6 +9,8 @@ use nemo_physical::{
     datatypes::data_value::DataValueIteratorT, dictionary::value_serializer::NULL_PREFIX,
 };
 
+use crate::io::parser::turtle::{decimal, double, integer};
+use crate::io::parser::{span_from_str, ParseError};
 use crate::model::{Identifier, NumericLiteral, RdfLiteral, Term};
 
 use super::{error::InvalidRuleTermConversion, primitive_types::PrimitiveType};
@@ -329,7 +331,6 @@ impl TryFrom<Term> for PhysicalString {
 
     fn try_from(term: Term) -> Result<Self, Self::Error> {
         match term {
-            // the first branch is the only one that results in an error
             Term::Variable(_) => Err(InvalidRuleTermConversion::new(term, PrimitiveType::Any)),
             Term::Constant(c) => Ok(c.into()),
             Term::NumericLiteral(NumericLiteral::Integer(i)) => Ok(LogicalInteger(i).into()),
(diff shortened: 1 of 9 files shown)


mmarx (changes requested):
Looks good, but could you maybe add a test for `"1000000000000000000000.0a"^^xsd:decimal` and `"1000000000000000000000.23"^^xsd:integer` being rejected?

mmarx:
Before this, we definitely rejected something like `"1a.0"^^xsd:decimal` (“Could not convert to `any`”), with this change, it seems we accept it.

monsterkrampe:
Yes, I hear you; I'm working on a solution :)
(We just rejected the really malformed ones because they were also not parsable as e.g. `i64`.)

monsterkrampe:
While it seems to work, I'm not happy with the solution yet. I feel like malformed literals like `"123.45a"^^xsd:integer` should not just be captured when converting a term to a physical string but already when creating the term.
I will try to shift things around a bit.

nemo/src/model/types/primitive_logical_value.rs
```diff
@@ -341,33 +343,126 @@ impl TryFrom<Term> for PhysicalString {
...
+                                Err(InvalidRuleTermConversion::new(term, PrimitiveType::Any))
+                            }
+                            _ => Ok(DatatypeValue(value.to_string(), datatype.to_string()).into()),
+                        }
+                    }
+                    _ => Err(InvalidRuleTermConversion::new(term, PrimitiveType::Any)),
+                },
+                XSD_DECIMAL => match nom::branch::alt((decimal, integer))(span_from_str(value)) {
+                    Ok((remainder, number)) => {
+                        eprintln!("{remainder:?};{number:?}");
```
mmarx, line 375: I guess those should still go (or, at the very least, become `log::debug!` invocations) ;)

nemo/src/model/types/primitive_logical_value.rs
```diff
@@ -341,33 +343,126 @@ impl TryFrom<Term> for PhysicalString {
...
+                        } else {
+                            match number {
+                                NumericLiteral::Integer(i) => Ok(Decimal(i, 0).into()),
+                                NumericLiteral::Decimal(a, b) => Ok(Decimal(a, b).into()),
+                                _ => unreachable!("We only parse as decimal or integer so only the above cases can occur.")
+                            }
+                        }
+                    }
+                    Err(nom::Err::Error(located_parse_error)) => {
+                        eprintln!("{located_parse_error:?}");
```
mmarx, line 390: see above

nemo/src/model/types/primitive_logical_value.rs
```diff
@@ -341,33 +343,126 @@ impl TryFrom<Term> for PhysicalString {
...
+                                    Ok(DatatypeValue(value.to_string(), datatype.to_string())
+                                        .into())
+                                }
+                            }
+                        }
+                        _ => Err(InvalidRuleTermConversion::new(
+                            term.clone(),
+                            PrimitiveType::Any,
+                        )),
                     }
```
mmarx, line 462: Note that `xsd:double` (and `xsd:float`) are distinct, finite datatypes with a value space disjoint from that of `xsd:decimal`. Not only is it perfectly fine to only accept what parses as `f64` (and `f32` for `xsd:float`, respectively), it's actually _required_.

mmarx (changes requested):
I'm not quite sure we want to invoke the parser here, since this is on the critical path for parsing. I would have assumed that it would be sufficient to check that all the characters in the literal (apart from, optionally, the sign and the decimal point) are digits, and this would also only be required in the fallback case.

nemo/src/model/types/primitive_logical_value.rs
monsterkrampe: Just forgot to remove those, thx :)

nemo/src/model/types/primitive_logical_value.rs
monsterkrampe: I see, that makes things simpler :)

monsterkrampe:
I get your concern with invoking the parser. I decided to not duplicate the logic and just use it. But since I only need to check integer and decimal anyway, this should be easy enough.
I will rewrite it.

monsterkrampe:
I decided to implement a (possibly failing) conversion from `RdfLiteral` to `Term` that is invoked when the `Term` is created.
This way, we are certain that the `Term`s are already normalized in the sense that literals of type e.g. `xsd:integer` have been converted to numeric literals (if possible).
This seems a bit cleaner to me than only doing this normalization when converting to physical strings. I hope that this does not affect performance too much.

mmarx (approved):

mmarx:
> This seems a bit cleaner to me than only doing this normalization when converting to physical strings. I hope that this does not affect performance too much.

I agree, this seems like the right thing to do.

PR #333, opened 2023-07-28
Unsafe negation

### LINKED ISSUE #330: Support Unsafe Variables in Negated Literals
Currently, the executing following rule would produce would cause a parse error:
```
h(?X) :- a(?X, ?Y), ~b(?Y, ?Z) .
```
This is because `?Z` does not appear in a positive atom. However, the intended meaning should be clear, namely that we consider an alternative negated atom `b'(?Y)` where the `?Z` column has been projected away.

Analogously, this should work:

```
h(?X) :- a(?X, ?Y), ~b(?Y, ?Z, ?Z) .
```
for expressing the condition that we want all `a(?X ?Y)` such that there is no `?Z` with `b(?Y, ?Z, ?Z)`. The underlying machinery for implementing this already exists for constants, so writing `~b(?Y, constant)` gets translated to `~b(?Y, ?FreshVariable)` and is handled correctly.

That being said, one should still disallow unsafe variables appearing in multiple negated atoms, like
```
h(?X) :- a(?X, ?Y), ~b(?Y, ?Z), ~c(?Z) .
```

aannleax:
Allows for unsafe variables in negated literals. For example:

```
result(?X, ?Y, ?Z) :- positive(?X, ?Y, ?Z), ~negative(?X, ?P, ?P) .
```

Also introduces a check that prevents rules that contain the same unsafe variable in multiple negated literals.

Closes #330.


nemo-physical/src/tabular/operations/triescan_nulls.rs
```diff
@@ -18,7 +18,7 @@ pub struct TrieScanNulls<'a> {
     /// Layer we are currently at in the resulting trie.
     current_layer: Option<usize>,
 
-    /// Types of the resulting trie.
+    /// Types of the resultingtodo!() trie.
```
mmarx, line 21: What happened here? :D

nemo-physical/src/tabular/operations/triescan_prune.rs
```diff
@@ -89,6 +89,15 @@ impl<'a> TrieScanPruneState<'a> {
         self.input_trie_scan_current_layer += 1;
     }
 
+    /// Returns the current layer of this trie scan
+    pub fn current_layer(&self) -> Option<usize> {
+        if self.initialized {
+            Some(self.external_current_layer)
+        } else {
+            None
+        }
```
mmarx, line 98: How about `self.initialized.then_some(self.external_current_layer)`?

nemo/src/model/rule_model/rule.rs
```diff
@@ -51,27 +51,34 @@ impl Rule {
...
+                }
+
+                current_unsafe.insert(variable.clone());
+            }
+
+            unsafe_negative_variables.extend(current_unsafe)
         }
 
+        // TODO: Implement something that checks whether unsafe negated variables occurs in multiple places
+
```
mmarx, line 81: This is what the code above does, isn't it?

mmarx (changes requested):
Just some very small nits to pick, looks good otherwise!

nemo-physical/src/tabular/operations/triescan_nulls.rs
aannleax: No clue :D

nemo/src/model/rule_model/rule.rs
aannleax: Yes, indeed

mmarx (approved):

PR #334, opened 2023-07-28
Add CLI option for input directory

### LINKED ISSUE #316: Directory for input files as a command line option
A command line option to define the input directory can be more comfortable than defining the input directory for each file in the rules file. This would be beneficial if somebody wants to run an analysis multiple times for different input directories without changing the code of the analysis.

monsterkrampe:
closes #316

### DIFF
--- a/nemo/src/io/resource_providers.rs
+++ b/nemo/src/io/resource_providers.rs
@@ -1,6 +1,6 @@
 //! Resource providers for external resources that can be used in reasoning.
 
-use std::{io::Read, rc::Rc};
+use std::{io::Read, path::PathBuf, rc::Rc};
 
 use flate2::read::MultiGzDecoder;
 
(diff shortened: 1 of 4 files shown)


nemo/src/io/resource_providers.rs
```diff
@@ -43,6 +43,14 @@ impl ResourceProviders {
         Self(Rc::new(r))
     }
 
+    /// Construct default with a base path for the `FileResourceProvider`
+    pub fn default_with_base_path(base_path: Option<PathBuf>) -> Self {
```
mmarx, line 47: I'd just call this `with_base_path`, since there's, e.g., `Vec::with_capacity` and no `Vec::default_with_capacity`.

mmarx (changes requested):

nemo/src/io/resource_providers.rs
monsterkrampe: Fine by me :)

mmarx (approved):

PR #342, opened 2023-08-16
add python support for more result types

matzemathics:


nemo-python/src/lib.rs
```diff
@@ -63,50 +66,116 @@ struct NemoOutputManager(nemo::io::OutputFileManager);
...
+        Python::with_gil(|py| {
+            let inner: String = value.extract(py).map_err(|_| {
+                NemoError::new_err(format!("Only string arguments are currently supported"))
+            })?;
+
+            Ok(NemoLiteral {
+                value: inner,
+                language: lang,
+                datatype: XSD_STRING.to_string(),
+            })
```
mmarx, line 99: This would allow the creation of a `NemoLiteral` with both a datatype and a language tag (which is not supported in either RDF nor in Nemo; in RDF, language-tagged strings always have `rdf:langString` as their implicit datatype). This is probably fine, since the bulk of those will be created from Nemo, but needs a bit if care if, at a later point, we convert a `NemoLiteral` into a primitive logical value.

nemo-python/src/lib.rs
```diff
@@ -63,50 +66,116 @@ struct NemoOutputManager(nemo::io::OutputFileManager);
...
+        })
+    }
+
+    fn value(&self) -> &str {
+        &self.value
+    }
+
+    fn datatype(&self) -> &str {
+        &self.datatype
+    }
```
mmarx, line 109: Do we need a getter for `language` as well?

nemo-python/src/lib.rs
```diff
@@ -63,50 +66,116 @@ struct NemoOutputManager(nemo::io::OutputFileManager);
...
+
+    fn __richcmp__(&self, other: &Self, op: pyo3::basic::CompareOp) -> PyResult<bool> {
+        match op {
+            pyo3::pyclass::CompareOp::Eq => Ok(self == other),
+            pyo3::pyclass::CompareOp::Ne => Ok(self != other),
+            _ => Err(PyNotImplementedError::new_err(
+                "rdf comparison is not implemented",
```
mmarx, line 116: Please capitalise `RDF`.

nemo-python/src/lib.rs
```diff
@@ -63,50 +66,116 @@ struct NemoOutputManager(nemo::io::OutputFileManager);
...
+                let lit = match lit {
+                    RdfLiteral::DatatypeValue { value, datatype } => NemoLiteral {
+                        value,
+                        language: None,
+                        datatype,
+                    },
+                    RdfLiteral::LanguageString { value, tag } => NemoLiteral {
+                        value,
+                        language: Some(tag),
+                        datatype: XSD_STRING.to_owned(),
```
mmarx, line 148: Language-tagged strings have datatype `rdf:langString`, cf. https://www.w3.org/TR/turtle/#sec-parsing-terms
> The literal has a lexical form of the first rule argument, String. If the '^^' iri rule matched, the datatype is iri and the literal has no language tag. If the LANGTAG rule matched, the datatype is rdf:langString and the language tag is LANGTAG. If neither matched, the datatype is xsd:string and the literal has no language tag.

mmarx (changes requested):
Two small changes and a question, looks good otherwise!

nemo-python/src/lib.rs
matzemathics: I don't think this is an issue, since this is only our internal representation of the python object that represents an rdf term. It can only be created by users via the `new()` method, which ensures the type of language strings is correctly set.

mmarx (approved):

PR #344, opened 2023-08-24
use a streaming permutator implementation

matzemathics:


nemo-physical/src/permutator.rs
```diff
@@ -212,6 +229,53 @@ impl Permutator {
     }
 }
 
+/// Iterates over data applying a permutation
+#[derive(Debug)]
+pub struct PermutatorStream<'a> {
+    index: usize,
+    permtation: &'a Permutator,
```
mmarx, line 236: `permutation` ;)

mmarx (changes requested):
Found a typo, looks good otherwise.

mmarx (approved):

Issue #355, opened 2023-09-13
Support equality of two integer variables


Laschoking:
For me this snipped does not work.
```
@declare foo1(integer).
@declare foo2(integer).
@declare intInBody(integer,integer).
foo1(1). foo2(2).
intInBody(?a,?b):-
    foo1(?a),
    foo2(?b),
    ?a = ?b.
```
Throwing the following error message:
```rust
thread 'main' panicked at nemo/src/execution/planning/plan_util.rs:35:39:
called `Option::unwrap()` on a `None` value
``` 
The other cases are working:
- inequality of two integer variables
- equality & inequality of one integer literals with one integer variables

matzemathics:
That is a bug in the normalization code: the head of the rule is not updated after normalization of the body.

PR #358, opened 2023-09-14
fix normalisation leading to invalid rules

matzemathics:
Fixes a bug discovered in #355 and generally tries to clean up the normalization code.


nemo/src/model/rule_model/aggregate.rs
```diff
@@ -7,6 +7,25 @@ pub struct Aggregate {
...
 
+impl Aggregate {
+    /// Substitutes all occurrences of `var` for `subst`.
+    pub fn substitute_variable(&mut self, var: &Variable, subst: &Variable) {
+        let Variable::Universal(var) = var else {
+            panic!("cannot substitute existential variable")
+        };
+
+        let Variable::Universal(subst) = subst else {
+            panic!("cannot substitute existential variable")
```
mmarx, line 18: maybe be slightly more explicit and have `cannot substitute with existential variable` here?

mmarx (changes requested):
Looks good, just have a typo and one panic that could be worded slightly more precisely.

mmarx (approved):

aannleax (approved):
Very nice

PR #362, opened 2023-09-19
Validate predicate usage against declared arity

### LINKED ISSUE #361: Arity of declare statement not checked against other uses of the atom
The following program runs with a runtime error but should produce a user error instead:
```
@declare B(integer, integer) .

A(?x, ?y) :- B(?x), B(?y) .
B(1) .
```

Expected: "error: Overloading of predicate names by arity is currently not supported."

mmarx:
Fixes #361.


aannleax (approved):
Nice, thanks

PR #363, opened 2023-09-19
Implement Tracing

aannleax:
This PR implements tracing, i.e. the functionality to output the origin of a derived fact. This can be done via the new command `--trace <Atom>`.

E.g. on program 
```
P(?x, ?y) :- Q(?y, ?x) .
S(?x) :- T(?x) .
R(?x, ?y) :- P(?x, ?y), S(?y) .

Q(a, b).
T(a). 
```
running `nmo --trace "R(b, a)"` produces 
```
 R(b, a) :- P(b, a), S(a) .
 ├─ P(b, a) :- Q(a, b) .
 │  └─ Q(a, b)
 └─ S(a) :- T(a) .
    └─ T(a)
```

Tracing is currently disabled for rules with arithmetic operations. Running the command on such rules will result in a user error. The reason for this is that this would require an implementation of arithmetic operators for filtering body matches.

### DIFF
--- a/nemo-cli/src/main.rs
+++ b/nemo-cli/src/main.rs
@@ -28,7 +28,11 @@ use colored::Colorize;
 use nemo::{
     error::{Error, ReadingError},
     execution::{DefaultExecutionEngine, ExecutionEngine},
-    io::{parser::parse_program, resource_providers::ResourceProviders, RecordWriter},
+    io::{
+        parser::{parse_fact, parse_program},
+        resource_providers::ResourceProviders,
+        RecordWriter,
+    },
     meta::{timing::TimedDisplay, TimedCode},
     model::OutputPredicateSelection,
 };
@@ -113,6 +117,12 @@ fn run(mut cli: CliApp) -> Result<(), Error> {
     log::info!("Rules parsed");
     log::trace!("{:?}", program);
 
+    let parsed_fact = if let Some(fact_string) = cli.trace_fact {
+        Some(parse_fact(fact_string)?)
+    } else {
+        None
+    };
+
     if cli.write_all_idb_predicates {
         program.force_output_predicate_selection(OutputPredicateSelection::AllIDBPredicates)
     }
@@ -184,6 +194,14 @@ fn run(mut cli: CliApp) -> Result<(), Error> {
 
     if cli.detailed_memory {
         println!("\n{}", engine.memory_usage());
+    }
+
+    if let Some(fact) = parsed_fact {
+        if let Some(trace) = engine.trace(fact.clone()) {
+            println!("\n{trace}");
+        } else {
+            println!("{fact} was not derived");
+        }
     }
 
     Ok(())
--- a/nemo-cli/src/main.rs
+++ b/nemo-cli/src/main.rs
@@ -197,7 +197,7 @@ fn run(mut cli: CliApp) -> Result<(), Error> {
     }
 
     if let Some(fact) = parsed_fact {
-        if let Some(trace) = engine.trace(fact.clone()) {
+        if let Some(trace) = engine.trace(fact.clone())? {
             println!("\n{trace}");
         } else {
             println!("{fact} was not derived");
--- a/nemo-cli/src/main.rs
+++ b/nemo-cli/src/main.rs
@@ -117,11 +117,7 @@ fn run(mut cli: CliApp) -> Result<(), Error> {
     log::info!("Rules parsed");
     log::trace!("{:?}", program);
 
-    let parsed_fact = if let Some(fact_string) = cli.trace_fact {
-        Some(parse_fact(fact_string)?)
-    } else {
-        None
-    };
+    let parsed_fact = cli.trace_fact.map(parse_fact).transpose()?;
 
     if cli.write_all_idb_predicates {
         program.force_output_predicate_selection(OutputPredicateSelection::AllIDBPredicates)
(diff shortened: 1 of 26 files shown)


nemo-cli/src/main.rs
```diff
@@ -113,6 +117,12 @@ fn run(mut cli: CliApp) -> Result<(), Error> {
     log::info!("Rules parsed");
     log::trace!("{:?}", program);
 
+    let parsed_fact = if let Some(fact_string) = cli.trace_fact {
+        Some(parse_fact(fact_string)?)
+    } else {
+        None
+    };
```
mmarx, line 124:
```rust
let parsed_fact = cli.trace_fact.map(parse_fact).transpose()?;
```

nemo-physical/src/management/database.rs
```diff
@@ -768,6 +769,31 @@ impl DatabaseInstance {
         }
     }
 
+    /// Evaluates a [`ExecutionTree`] until until it finds the first row in the result and returns it.
```
mmarx, line 772:
```
Evaluates an [`ExecutionTree`]
```

nemo-physical/src/management/database.rs
```diff
@@ -768,6 +769,31 @@ impl DatabaseInstance {
...
+    /// Returns `None` if the tree evaluates to the empty table.
+    ///
+    /// # Panics
+    /// Panics if the tables that are being loaded by the [`ExecutionTree`] are not available in memory.
+    /// Also panics if the [`ExecutionTree`] wants to perform a project/reorder operation on a non-materialized trie.
+    fn evaluate_execution_tree_until_first(
```
mmarx, line 778: We have `until_first` here and `first_match` down below, since they mean the same thing, they should have a uniform name (`until_first_match`, e.g.).

nemo-physical/src/tabular/operations/materialize.rs
```diff
@@ -72,17 +75,29 @@ pub fn materialize(trie_scan: &mut impl TrieScan) -> Option<Trie> {
...
 }
 
-/// Tests whether an iterator is empty by materializing it until the first element
-pub fn scan_is_empty(trie_scan: &mut impl TrieScan) -> bool {
-    trie_scan
-        .advance_on_layer(trie_scan.column_types().len())
-        .is_some()
+/// Tests whether an iterator is empty by materializing it until the first element.
+/// Returns the first row of the result or `None` if it is empty.
+pub fn scan_is_empty(trie_scan: &mut impl TrieScan) -> Option<TableRow> {
```
mmarx, line 80: Since this no longer returns boolean, the name should change to reflect that, maybe `scan_first_match`?

nemo/src/io/parser.rs
```diff
@@ -35,6 +35,18 @@ pub fn parse_program(input: impl AsRef<str>) -> Result<Program, Error> {
...
+///
+/// The program will be parsed and checked for unsupported features.
+///
+/// # Error
+/// Returns an appropriate [`Error`] variant on parsing and feature check issues.
+pub fn parse_fact(mut input: String) -> Result<Fact, Error> {
+    input += ".";
+    let fact = all_input_consumed(RuleParser::new().parse_fact())(input.as_str())?;
+    Ok(fact)
+}
```
mmarx, line 48: This is fine in principle, but using a fresh parser has the drawback that the fact cannot use, e.g., prefixed names that would need resolving.

nemo/src/model/rule_model/term.rs
```diff
@@ -105,11 +105,32 @@ impl Variable {
             Self::Universal(identifier) | Self::Existential(identifier) => identifier.name(),
         }
     }
+
+    /// Return whether this is a universal variable.
+    pub fn is_universal(&self) -> bool {
+        match self {
+            Variable::Universal(_) => true,
+            Variable::Existential(_) => false,
+        }
```
mmarx, line 114:
```rust
matches!(self, Variable::Universal(_))
```

nemo/src/model/rule_model/term.rs
```diff
@@ -105,11 +105,32 @@ impl Variable {
...
+    }
+
+    /// Return whether this is an existential variable.
+    pub fn is_existential(&self) -> bool {
+        match self {
+            Variable::Universal(_) => false,
+            Variable::Existential(_) => true,
+        }
```
mmarx, line 122:
```rust
matches!(self, Variable::Existential(_))
```

nemo/src/execution/execution_engine.rs
```diff
@@ -378,4 +406,255 @@ impl<Strategy: RuleSelectionStrategy> ExecutionEngine<Strategy> {
...
+                            entry.insert(fact_term.clone());
+                        }
+                    }
+                } else {
+                    unreachable!("Variables are the only non-ground terms");
+                }
+            }
+
+            if !compatible {
+                // Fact could not haven unified
```
mmarx, line 501: Not sure what this should say (`could not be unified`, maybe?), but `haven` is surely not the right word here.

mmarx (changes requested):
Looks good, just found a few typos and some opportunities for simpler code.

nemo/src/io/parser.rs
aannleax: Yeah its a pretty lazy solution.

Maybe `parse_program` should return the `RuleParser` as well? Then I could use it to parse the additional fact

nemo/src/execution/execution_engine.rs
aannleax: Had a little stroke there or something

mmarx (approved):

PR #365, opened 2023-09-20
Add basic aggregate support

rlwww:


nemo/src/model/types/error.rs
```diff
@@ -28,6 +28,9 @@ impl From<InvalidRuleTermConversion> for ReadingError {
 /// Errors that can occur during type checking
 #[derive(Error, Debug)]
 pub enum TypeError {
+    /// Non-numerical aggregate input type
+    #[error("Aggregate operation \"{0:?}\" on input variable \"{1}\" requires a numerical input type, but was input type was \"{2}\".")]
```
mmarx, line 32: `but was input type was` ;)

nemo/src/model/chase_model/aggregate.rs
```diff
@@ -0,0 +1,35 @@
...
+    pub(crate) aggregate_operation: AggregateOperation,
+    pub(crate) logical_aggregate_operation: LogicalAggregateOperation,
+
+    pub(crate) variable_identifiers: Vec<Identifier>,
+    /// Name of placeholder variable for the aggregate, which will get used in the corresponding head atom.
+    pub(crate) output_variable: Identifier,
+}
+
+impl ChaseAggregate {
+    #[allow(missing_docs)]
```
mmarx, line 17: Why no docs here?

mmarx (changes requested):
Looks good, just a few typos and some questions.

nemo-physical/src/aggregates/processors/processor.rs
```diff
@@ -0,0 +1,29 @@
...
+/// Allows aggregation of multiple rows (all with the same group-by values) to produce a single aggregate value.
+pub trait AggregateGroupProcessor<A>
+where
+    A: Aggregate,
+{
+    /// Processes a row of the aggregated input column and updates the internal state.
+    fn write_aggregate_input_value(&mut self, value: A);
+
+    /// Returns the resulting aggregated value of all the processed input values.
+    fn finish(&self) -> Option<StorageValueT>;
```
monsterkrampe, line 28: General Question: Shouldn't the aggregates work with `DataValueT`, or put differently: Do we not want to support aggregates on strings?

mmarx (approved):

monsterkrampe (approved):

nemo-physical/src/aggregates/processors/processor.rs
monsterkrampe: I'm aware that it's not easy to talk about `DataValueT` in a `TrieScan`. Intuitively, I feel that aggregation should happen on a higher level (but performance will suffer from that). Anyway, for now it should be fine :)

aannleax (changes requested):
Very nice work :)

Issue #360, opened 2023-09-18
Refactor `Term` model


matzemathics:
currently the `Term` type contains a lot of different things, which do not necessarily play nicely together:
- it is used as the type of values, that are returned from the logical layer when iterating through a table
- yet it also contains variables and even aggregates, which do not make sense in the above context
- however it does _not_ contain arithmetic operations, which are encoded using the separate `TermTree` type (even though they too could reosonably referred to as "terms")

I propose the following model:

```rust
enum Term {
  /// An arithmetic operation, aggregation or function call
  Operation {
    /// The operation to be performed
    operation: TermOperation,
    /// The arguments to the operation
    subterms: Vec<Term>
  },
  /// A (universal or existential) variable
  Variable(Variable),
  /// A constant of type primitive
  Constant(PrimitveValue),
}

enum PrimitiveValue {
  /// An (abstract) constant.
  Constant(Identifier),
  /// A numeric literal.
  NumericLiteral(NumericLiteral),
  /// A string literal.
  StringLiteral(String),
  /// An RDF literal.
  RdfLiteral(RdfLiteral),
}
```

Notably, I would propose to make aggregations a special case of `TermOperation` instead of a seperate thing. @rlwww do you see any problems with this?
Another thing to keep in mind is that while we only support primitive constants right now, the API that returns values from tables will need something to represent the `Any` type, so something like this might be added:

```rust
enum AnyValue {
  /// A value of type primitive
  Primitive(PrimitiveConstant),
  // other complex values...
}
```

That way there is a clean seperation between a `Term`, which is a parsed representation of user input, and a computed `AnyValue`, which is what comes out at the end of reasoning and can be serialized into various formats.

matzemathics:
I have looked at the definition of aggregate again, since it currently only supports variables as subterms, it might make sense to have a special case for that.

aannleax:
I like the overall idea, but I don't get the point behind the additional `AnyValue` structure to separate "user input" from "computed values". Why shouldn't "`// other complex values...`" be written by the user? Is "Any" the only exception (although, if we allow implicit type declarations via constants (like writing `A(1, 2.0)` to imply `@declare A(integer, float64)`), it's not unreasonable to allow something like `B(Any::Integer(1))` and have it imply `@declare B(any)`., so I'm not even sure if it should be an exception)?

mkroetzsch:
[...]

PR #371, opened 2023-10-09
Feature/366 body arithmetic

aannleax:
The main addition is the implementation of complex arithmetic to filter body matches, i.e.
```
complex(?X, ?Y) :- data(?X, ?Y), Abs(?X - ?Y) >= ?X * ?X.
```
This also includes support for new arithmetic operations (Absolute value and square root; also the physical layer support unary minus and exponentiation but someone needs to tell this to the parser).

Furthermore, this introduces a new operator `:=`, which can be used to define a variable like so:
```
r(?R, 5, ?R, ?S) :- a(?X, ?Y, ?T), b(?Z, ?X), ?R := (?X + 3) * ?Z, ?S := ?Y - 1.
```
The reason for this is to have simple rules for then a variable is safe. Now this would be if that variable either occurs as a simple term in a positive body literal or is defined via `:=`. We could, in the future, have it so expressions like `A(?x), ?x = ?y + 5` could define a variable (in this case `?y`). But this is perhaps a bit messy ... At least some further discussion is needed on this.

Since these changes required messing with the rule model, I took the opportunity to refactor a lot of things in the rule and chase rule model. Our object model no longer uses `Term`s and `TermTree`s but, in spirit of #360, defines `Constant`, which can be be built into `PrimitiveTerm`s, which are the leaf node in a complex term tree, which is now simply named `Term`. The chase model has been altered to have less degrees of freedom in its representation.

Also, I broke parsing for nested functions.


nemo-physical/src/columnar/operations/arithmetic/expression.rs
```diff
@@ -0,0 +1,471 @@
+//! Implements the functionality to represent and evaluate complex arithmetic expressions
+
+use std::fmt::{Debug, Display};
+
+use super::traits::ArithmeticOperations;
+
+/// Tree representing an arithmetic expression
+#[derive(Clone)]
+pub enum ArithmeticTree<T> {
```
matzemathics, line 9: I'd imagine, it would be helpful for this enum to be more structural e.g.
```rust
enum ArithmeticTree<> {
  Constant(..),
  Reference(..),
  BinaryOperation(..),
  UnaryOperation(..)
}
```

Probably postpone it though, since this tree structure should live in the nemo module, not in the physical layer.

nemo/src/execution/planning/plan_util.rs
```diff
@@ -34,170 +35,54 @@ use crate::{
...
+        let variable_type = variable_types
+            .get(some_variable)
+            .expect("Every variable should have a type.");
+
+        let (left, right) = constraint.terms();
+        let left_tree = termtree_to_arithmetictree(left, variable_order, variable_type);
+        let right_tree = termtree_to_arithmetictree(right, variable_order, variable_type);
+
+        let new_statement = match constraint.condition() {
+            Condition::Assignment(_, _) => unreachable!("Not a valid constraint"),
```
matzemathics, line 71: "Not a valid constraint" seems a bit broad. Something along the lines of "Equality constraints are normalized into joins." (if this is even correct)?

nemo/src/model/rule_model/rule.rs
```diff
@@ -11,158 +11,156 @@ pub struct Rule {
...
+            .collect::<HashSet<_>>();
 
-            for variable in negative_literal.variables() {
-                if positive_varibales.contains(variable) {
-                    continue;
+        // A derived variable is a variable which is defined in terms of bound variables
+        let mut derived_variables = HashSet::<&Variable>::new();
+        for condition in &conditions {
+            if let Condition::Assignment(variable, term) = condition {
+                if bound_variables.contains(variable) || derived_variables.contains(variable) {
```
matzemathics, line 79: Is it totally unreasonable to have more than one equality constraint on one variable? I guess something like
`p(?prev, ?x, ?next) :- q(?prev), q(?x), q(?next), ?x = ?prev - 1, ?x = ?next + 1.`
might be desirable.

nemo/src/model/rule_model/term.rs
```diff
@@ -128,204 +43,393 @@ impl std::fmt::Display for Variable {
...
+    pub name: Identifier,
+    /// Its subterms
+    pub subterms: Vec<Term>,
 }
 
-impl std::fmt::Display for NumericLiteral {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+/// Possibly complex term that may occur within an [`super::Atom`]
+#[derive(Eq, PartialEq, Clone, PartialOrd, Ord)]
+pub enum Term {
```
matzemathics, line 125: Again here I would strongly prefer something more structural like commented above, because otherwise I believe we risk spreading very long match statements with repetitive code.

matzemathics (changes requested):

nemo-physical/src/columnar/operations/arithmetic/expression.rs
aannleax: Yeah, I mean this will get replaced anyways. But good idea nonetheless.

nemo/src/execution/planning/plan_util.rs
aannleax: Not quite. Basically, a "constraint" is the same as a "condition" except for "Assignment" (which works slightly different from equality). I just laziliy reused the same structure instead of creating another enum with 90% of the same values.

I guess a nicer way to do this would be to define Constraints and then have 
```rs
enum Condition {
   Constraint(Constraint),
   Assignment(var, term)
}
```

nemo/src/model/rule_model/term.rs
aannleax: Yeah I can see that. I'll try to accommodate this

nemo/src/model/rule_model/rule.rs
aannleax: Assignment is another operator `:=` which works different from equality. You can have multiple equality restrictions on one variable

matzemathics (approved):
I'd say this looks good now.

mmarx (changes requested):
I have some small (mostly stylistic) changes, and an insight into why this breaks nested functions.

nemo-physical/src/management/type_analysis.rs
```diff
@@ -243,7 +243,9 @@ impl TypeTree {
 
                 Ok(TypeTreeNode::new(new_schema, vec![subtype_node]))
             }
-            ExecutionOperation::SelectValue(subtree, _assignments) => {
+            ExecutionOperation::Filter(subtree, _condition) => {
+                // TODO: Shouldnt I do something here?
```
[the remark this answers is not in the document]
aannleax: Filter expressions do not impose type constraints. Hence I shouldn't, in fact be doing anything here. It just seemed odd to me when I encountered this part of the code and wanted to think about it later.

So I removed the todo again

nemo/src/io/parser.rs
```diff
@@ -982,54 +1039,53 @@ impl<'a> RuleParser<'a> {
...
             map_error(
                 move |input| {
                     delimited(
                         multispace_or_comment0,
                         alt((
-                            self.parse_function_term(),
                             self.parse_arithmetic_expression(),
-                            self.parse_parenthesised_term_tree(),
+                            self.parse_parenthesised_term(),
+                            self.parse_function_term(),
```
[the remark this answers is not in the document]
aannleax: Should be fixed now

aannleax:
As was discussed, I removed the `:=` operator. The same functionality can now be achieved by using the regular `=` operator.

mmarx (approved):
Looks good!

PR #372, opened 2023-10-11
Respect literals (and facts) when inferring types

### LINKED ISSUE #359: Type inference does not catch / coerce mismatch, leading to bug in the backend.
Following code panics in debug mode, release mode would invoke undefined behavior.
```
@declare p(integer, integer).
a(1). b(2).
p(?x,?y) :- a(?x), b(?y).
```
Here's the output:
```
thread '<unnamed>' panicked at nemo/src/model/types/primitive_logical_value.rs:539:18:
internal error: entered unreachable code: If the database representation of the logical types is correct, we never reach this branch.
```

monsterkrampe:
It should fix #359 ~but still needs to be cleaned up, so no need to review yet.~ I will turn in into a normal PR once I'm somewhat confident :)

EDIT: **This is ready now.**  
There is still some inferences that are not done but probably should be. This concerns nested terms and especially nested variables in constraints (formerly filters). It's not unlikely that I overlooked something else. Nevertheless, since this is only a bugfix, I would keep this as it is for now. I assume that this will change with further advances in the type system anyway.  
I plan to write up a wiki page with the current behavior of the type inference code. There we can also discuss the intended behavior for the future.


nemo/src/program_analysis/analysis.rs
```diff
@@ -660,6 +660,92 @@ impl ChaseProgram {
                 })
                 .collect::<HashMap<_, _>>();
 
+            let mut fact_and_constant_term_decls: HashMap<Identifier, Vec<Option<PrimitiveType>>> =
```
matzemathics, line 663: This whole function was already very long and hard to follow. It may even deserve its own module. Would you think it is reasonable to split this thing off into a new module and break it into comprehensible pieces?

nemo/src/program_analysis/analysis.rs
```diff
@@ -660,6 +660,92 @@ impl ChaseProgram {
...
+                                .iter()
+                                .zip(new_types.iter())
+                                .map(|(a, b)| {
+                                    a.as_ref()
+                                        .map(|a| b.as_ref().map(|b| a.max_type(b)).unwrap_or(*a))
+                                        .or(*b)
+                                })
+                                .collect();
+                        })
+                        .or_insert(new_types);
```
matzemathics, line 745: If this pattern is common, maybe it deserves its own function?

nemo/src/program_analysis/analysis.rs
```diff
@@ -660,6 +660,92 @@ impl ChaseProgram {
...
+                            .iter()
+                            .zip(new_types.iter())
+                            .map(|(a, b)| {
+                                a.as_ref()
+                                    .map(|a| b.as_ref().map(|b| a.max_type(b)).unwrap_or(*a))
+                                    .or(*b)
+                            })
+                            .collect();
+                    })
+                    .or_insert(new_types);
```
matzemathics, line 686: I'm not following. Where does `types` vs. `new_types` come from? Why do we take the max?

matzemathics (changes requested):
Overall this approach looks very reasonable, the changes to the model look fine. However I think we should not further complicate the logic in `analysis.rs`, because it gets into the territory where it is hard to comprehend and make changes to.

nemo/src/model/rule_model/numeric_literal.rs
```diff
@@ -11,6 +13,17 @@ pub enum NumericLiteral {
     Double(Double),
 }
 
+impl NumericLiteral {
+    /// Get primitive type that fits the literal
+    pub fn primitive_type(&self) -> PrimitiveType {
+        match self {
+            Self::Integer(_) => PrimitiveType::Integer,
+            Self::Double(_) => PrimitiveType::Float64,
+            Self::Decimal(_, _) => PrimitiveType::Float64, // TODO: is this what we want?
```
mmarx, line 22: Pretty sure this is not what we want, since `f64` is not a faithful representation of `Decimal`. In particular, there are integers in `u64` (and hence, also in `Decimal`) that are not representable in `f64`.

nemo/src/model/types/primitive_types.rs
```diff
@@ -130,6 +130,30 @@ impl TryFrom<NestedType> for PrimitiveType {
...
+    pub fn max_type(&self, other: &PrimitiveType) -> PrimitiveType {
+        match self {
+            PrimitiveType::Any => PrimitiveType::Any,
+            PrimitiveType::String => match other {
+                PrimitiveType::String => PrimitiveType::String,
+                _ => PrimitiveType::Any,
+            },
+            PrimitiveType::Integer => match other {
+                PrimitiveType::Integer => PrimitiveType::Integer,
+                _ => PrimitiveType::Any,
```
mmarx, line 148: I think we should handle the case where we have `Integer` and `Decimal` here (which should promote to `Decimal`, _not_ to `Any`).

nemo/src/model/rule_model/numeric_literal.rs
monsterkrampe: True, this should probably fall back to `Any` then...

nemo/src/model/types/primitive_types.rs
monsterkrampe: There is no `Decimal` variant in `PrimitiveType`.

monsterkrampe:
Thanks for the reviews!
Refactorings are certainly necessary, I just did not see them in scope of the bugfix...
I will invest the afternoon and see what I can do :)

nemo/src/model/types/primitive_types.rs
mmarx: Ah, right.

nemo/src/program_analysis/analysis.rs
```diff
@@ -1041,537 +387,41 @@ impl ChaseProgram {
...
-        let predicate_types = self.infer_predicate_types(&position_graph, &all_predicates)?;
+        let (predicate_types, rule_var_types) = infer_types(self)?;
 
         let rule_analysis: Vec<RuleAnalysis> = self
             .rules()
             .iter()
+            .zip(rule_var_types)
             .enumerate()
-            .map(|(i, r)| {
+            .map(|(i, (r, vt))| {
```
matzemathics, line 397: `(index, (rule, variable_types))` ?

monsterkrampe:
Turned out to be a bit more time than just the afternoon but I think that this is ready for another review cycle now :)

nemo/src/program_analysis/type_inference/position_graph.rs
```diff
@@ -0,0 +1,319 @@
...
+            for atom in rule.all_body() {
+                for (term_position, variable) in atom.terms().iter().enumerate() {
+                    let predicate_position =
+                        PredicatePosition::new(atom.predicate(), term_position);
+
+                    // Add head position edges
+                    {
+                        // NOTE: we connect each body position to each head position of the same variable
+                        if let Some(head_positions) = variables_to_head_positions.get(variable) {
+                            for pos in head_positions {
```
matzemathics, line 99: Just an idea: maybe this can be written as a single loop using `flatten()`, to remove one layer of indentation like so:
```rust
let head_positions = variables_to_head_positions.get(variable);
for position in head_positions.into_iter().flatten() {
  ..
}
```

nemo/src/program_analysis/type_inference/position_graph.rs
```diff
@@ -0,0 +1,319 @@
...
+                                );
+                            }
+                        }
+
+                        // NOTE: we connect every aggregate input variable to it's corresponding output variable in the head
+                        if let Some(output_variable_identifiers) =
+                            aggregate_input_to_output_variables.get(variable)
+                        {
+                            for (output_variable_identifier, edge_label) in
+                                output_variable_identifiers.iter()
```
matzemathics, line 113: same pattern here

nemo/src/program_analysis/type_inference/type_requirement.rs
```diff
@@ -0,0 +1,308 @@
...
+        for chase_atom in chase_rule
+            .all_body()
+            .cloned()
+            .map(Into::into)
+            .chain(chase_rule.head().iter().cloned())
+        {
+            let reqs_for_atom: Vec<TypeRequirement> = chase_atom
+                .terms()
+                .iter()
+                .map(|t| match t {
```
matzemathics, line 213: `t` -> term or type?

matzemathics: *I guess term

matzemathics (approved):
Overall looks like a very solid refactor. I have some small style nits, but otherwise I think this is fine.

nemo/src/model/rule_model/numeric_literal.rs
```diff
@@ -11,6 +13,19 @@ pub enum NumericLiteral {
...
 }
 
+impl NumericLiteral {
+    /// Get primitive type that fits the literal
+    pub fn primitive_type(&self) -> PrimitiveType {
+        match self {
+            Self::Integer(_) => PrimitiveType::Integer,
+            Self::Double(_) => PrimitiveType::Float64,
+            Self::Decimal(_, 0) => PrimitiveType::Integer,
+            // this is a conservative estimate; if possible we transform decimals into Integers or Float64 but if this does not work, we store it as Any (internally as a String)
```
matzemathics, line 23: I guess technically we never want to convert decimals into float (at least not implicitly), because floating point arithmetic gives different results than working on decimals.

nemo/src/program_analysis/type_inference/position_graph.rs
monsterkrampe: Ah yes, I never really considered the possibility of turning an `Option` into an iterator but this makes a lot of sense :)

nemo/src/model/rule_model/numeric_literal.rs
monsterkrampe: One can certainly discuss this. As of now however, there is the following implementation:
```
impl TryFrom<Constant> for LogicalFloat64 {
    type Error = InvalidRuleTermConversion;

    fn try_from(constant: Constant) -> Result<Self, Self::Error> {
        match constant {
            Constant::NumericLiteral(NumericLiteral::Double(d)) => Ok(d.into()),
            Constant::NumericLiteral(NumericLiteral::Decimal(a, b)) => format!("{a}.{b}")
                .parse()
                .ok()
                .and_then(|d: f64| Double::new(d).map(|d| d.into()).ok())
                .ok_or(InvalidRuleTermConversion::new(
                    constant,
                    PrimitiveType::Float64,
                )),
            Constant::NumericLiteral(NumericLiteral::Integer(a)) => LogicalInteger(a)
                .try_into()
                .map_err(|_err| InvalidRuleTermConversion::new(constant, PrimitiveType::Float64)),
            _ => Err(InvalidRuleTermConversion::new(
                constant,
                PrimitiveType::Float64,
            )),
        }
    }
}
```
But it might be that this never comes into action with the dataflow we have.

mmarx (approved):

PR #380, opened 2023-10-18
fix skipping over values in constraint implementation

matzemathics:


matzemathics:
@aannleax I'm unsure whether there could be a similar problem in line 380.
https://github.com/knowsys/nemo/pull/380/files#diff-49d351e3f72e81729798365883f856f6de2870555aeac5fd421d34a3cfca0591L308

aannleax:
I'm not sure what we want actually. For something like `R(?z) :- A(?x, ?y), ?z = ?x / ?y`, we want to skip all results with `?y = 0`. For constraints `R(?x, ?y) :- A(?x, ?y), ?x / ?y > 10` I'm not sure but I think we also want to skip the rows (so treat it as a if the constraint was not satisfied).

To do this the function `check_condition` should probably just return false if the condition is not satisfied or if the expression contains an undefined expression.

`satisfy_lower_bound` should probably return nothing and loop while tree.evaluate returns None

`check_upper_bound` could stay as before, but the code that uses it should probably loop  until it returns either true or false

aannleax:
I can also do the fixes, if you want

matzemathics:
> I can also do the fixes, if you want

Yes feel free to pick this up, I'm not sure I understand all your remarks.

Skipping rows, where division by zero occurs is expected for me, the sole problem is that it also skipped non-divide-by-zero-rows that are in the same group as a divide-by-zero-row, which seems unexpected to me.

aannleax:
Ok, I think it makes sense now

matzemathics:
> Ok, I think it makes sense now

Nice, I will try and add some more test cases then we can get this merged

nemo-physical/src/columnar/operations/columnscan_restrict_values.rs
```diff
@@ -234,44 +246,64 @@ where
...
+    /// Returns `true` if all conditions are satisfied.
+    fn check_conditions(&self, referenced_values: &[T]) -> bool {
         for condition in &self.conditions {
-            if !condition.evaluate(referenced_values)? {
-                return Some(false);
+            if let Some(result) = condition.evaluate(referenced_values) {
+                if result {
+                    continue;
+                }
             }
```
mmarx, line 301:
```rust
if let Some(true) = condition.evaluate(referenced_values) {
  continue;
}
```

matzemathics:
@aannleax @mmarx anything still missing? otherwise I'd suggest we get this in.

aannleax (approved):

PR #387, opened 2023-10-22
refactor term representation in the physical layer

matzemathics:


mmarx (approved):

matzemathics:
rebased on top of #372, so should not be merged before that

PR #388, opened 2023-10-23
Drop `operation` and `seek_iai` benches

mmarx:
* `iai` is unmaintained and breaks with `valgrind-3.21`. The `seek_iai` bench exercises the same code as the `seek` bench.
* the `operations` bench pulls in polars, which is a huge dependency that regularly breaks its API on (minor) version bumps. Furthermore, the bench itself uses internal parts from the `nemo-physical` API and thus also regularly requires changes, often only caught in CI, since the bench runs impractically long to actually be useful (still not finished after an hour on Wille – edit: failed after 1:56:00 since `test-files/bench-data/aux-split/aux-0.csv` was not available), and the exercised code is now covered by the end-to-end tests. It also does not help that non-redistributable test data is required for the bench to successfully execute.


matzemathics:
> edit: failed after 1:56:00 since test-files/bench-data/aux-split/aux-0.csv was not available

haha, how long has it been since someone ran that?

mmarx:
> haha, how long has it been since someone ran that?

too long, but that was basically my fault for not uncompressing the file. In any case, two hours is _way_ too long for any single bench.

matzemathics:
Finally that should be rectified when #311 is implemented.

PR #389, opened 2023-10-23
Don't run miri on slow unit tests

mmarx:
This brings down the total run time for `cargo miri test` by over 10 minutes on my machine.


mmarx:
Rebased on top of #388 for the `clippy` fixes (so should not be merged before that). Notice how the `miri` check completed in 6 minutes versus 32 minutes for #388.

PR #390, opened 2023-10-24
Downgrade bytecount to 0.6.4

mmarx:
Bytecount 0.6.5 is currently broken on WASM, see
https://github.com/llogiq/bytecount/issues/89.

### DIFF
--- a/nemo/Cargo.toml
+++ b/nemo/Cargo.toml
@@ -30,6 +30,7 @@ thiserror = "1.0"
 flate2 = "1"
 sanitise-file-name = "1.0.0"
 nom_locate = { version = "4.1.0", features = [ "runtime-dispatch-simd" ] }
+# Bytecount 0.6.5 is currently broken on WASM, see llogiq/bytecount#89.
 bytecount = "=0.6.4"
 getrandom = { version = "0.2.9", default-features = false }
 path-slash = "0.2.1"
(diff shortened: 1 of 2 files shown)


matzemathics:
@mmarx shall we pin it in the cargo.toml?

mmarx:
Not sure if that's necessary, it's only pulled in as a transitive dependency, so having it in the `Cargo.lock` should suffice.

matzemathics:
i guess it would break when doing cargo update / not sure about cargo add. But we can of course also not do that.

mmarx:
Nevermind, we need to pin it since otherwise the nix build tries to rewrite `Cargo.lock`, which fails due to the sandboxing.

nemo/Cargo.toml
```diff
@@ -30,6 +30,7 @@ thiserror = "1.0"
 flate2 = "1"
 sanitise-file-name = "1.0.0"
 nom_locate = { version = "4.1.0", features = [ "runtime-dispatch-simd" ] }
+bytecount = "=0.6.4"
```
matzemathics, line 33: ah, I would have suggested to do a patch version instead

mmarx: This is really just a temporary workaround, I don't want to keep this around once a fixed version is available.

matzemathics: nvm, patches must point to a different location

matzemathics (approved):

mmarx:
Ah, the comment is a good idea, thanks!

PR #392, opened 2023-10-27
Fix bug with unknown constants in filter expressions

### LINKED ISSUE #391: Panic when filtering for non-existing constant
On branch main, commit 3181377

Consider the following minimal example
```
p("a", "b") .
q(?X) :- p(?X, "c") .
```

nemo panics with the following error
```
thread 'main' panicked at nemo-physical/src/tabular/operations/triescan_select.rs:347:41:
We don't have string operations so this cannot fail.
```

aannleax:
Should fix #391.


mmarx (approved):
LGTM

PR #393, opened 2023-10-27
patch: shortening primitive logical value prefixes

larry-gonzalez:
Currently primitive logical values uses prefixes to distinguish among data types
Prefixes lengths range from 7 to 16 characters, which uses unnecessary memory

In this (patch) pull request we shorten all all prefixes to 3 characters and update tests accordingly

Experiments show memory savings up to ~300MB when loading 100M triples from Wikidata


monsterkrampe (approved):
Fine by me :)

PR #395, opened 2023-11-01
Fix type compatibility where casting is needed

### LINKED ISSUE #394: Attempt to upgrade values to `any` leads to unreachable code being executed.
Example:
```prolog
values(42).

@declare result(any).
result(?v) :- values(?v).
```
The type inference (correctly) infers the type of `values` should be `integer`. However it does not complain, that there is currently no support for casting integers into `any`, eventually reaching unreachable code.
```
thread '<unnamed>' panicked at nemo/src/model/types/primitive_logical_value.rs:407:18:
internal error: entered unreachable code: If the database representation of the logical types is correct, we never reach this branch.
```

monsterkrampe:
Fixes #394

### DIFF
--- a/nemo/src/io/parser.rs
+++ b/nemo/src/io/parser.rs
@@ -472,7 +472,7 @@ impl<'a> RuleParser<'a> {
                 move |input| {
                     let (remainder, (predicate, tuple_constraint)) = preceded(
                         terminated(token("@source"), cut(multispace_or_comment1)),
-                        cut(self.parse_qualified_predicate_name()),
+                        cut(self.parse_qualified_predicate_name(true)),
                     )(input)?;
 
                     let (remainder, datasource): (_, Result<_, ParseError>) = cut(delimited(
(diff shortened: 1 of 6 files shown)


nemo/src/model/types/complex_types.rs
```diff
@@ -133,14 +133,6 @@ impl Deref for TupleConstraint {
     }
 }
 
-impl FromIterator<PrimitiveType> for TupleConstraint {
```
monsterkrampe, line 136: I removed this since I found it confusing. I was actually expecting that this would always be turned into `AtLeast` constraints instead of `Exact` constraints, e.g. when using this on Data Sources.
I think one should rather make this explicit and use the `FromIterator` implementation below.

nemo/src/io/parser.rs
```diff
@@ -690,6 +690,7 @@ impl<'a> RuleParser<'a> {
     /// predicate name together with its arity.
     fn parse_qualified_predicate_name(
         &'a self,
+        constraint_is_lower_bound: bool,
```
monsterkrampe, line 693: When parsing source declarations, the type annotations are always lower bounds. In other cases like outputs, these requirements might be stricter. (Since import / export directives are WIP, I would not polish this right now as I expect things to change in the parser anyway.)

mmarx (approved):

PR #400, opened 2023-11-06
Expose version info in nemo-wasm and add cargo config

rlwww:


mmarx (approved):

mmarx:
Looks good, but note that the recent dictionary merge broke the wasm build :/

PR #402, opened 2023-11-07
add wasm-pack test on pr

matzemathics:
see: https://rustwasm.github.io/docs/wasm-bindgen/wasm-bindgen-test/continuous-integration.html#github-actions. note that currently no tests are implemented for nemo-wasm.

### DIFF
--- a/nemo-wasm/Cargo.toml
+++ b/nemo-wasm/Cargo.toml
@@ -20,5 +20,7 @@ nemo-physical = { path = "../nemo-physical", default-features = false }
 thiserror = "1.0"
 wasm-bindgen = "0.2"
 wasm-bindgen-futures = "0.4.37"
-wasm-bindgen-test = "0.2"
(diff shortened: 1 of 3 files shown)


nemo-wasm/Cargo.toml
```diff
@@ -20,4 +20,5 @@ nemo-physical = { path = "../nemo-physical", default-features = false }
 thiserror = "1.0"
 wasm-bindgen = "0.2"
 wasm-bindgen-futures = "0.4.37"
+wasm-bindgen-test = "0.2"
```
mmarx, line 23: shouldn't this be under `[dev-dependencies]`?

mmarx (changes requested):

nemo-wasm/Cargo.toml
matzemathics: ah right.

mmarx (approved):

PR #407, opened 2023-11-08
Non-experimental aggregates

rlwww:


nemo/src/execution/planning/aggregates.rs
```diff
@@ -0,0 +1,212 @@
...
+            AggregationInstructions {
+                aggregate_operation: aggregate.aggregate_operation,
+                group_by_column_count: aggregate_group_by_variables.len(),
+                aggregated_column_index,
+                last_distinct_column_index,
+            },
+        );
+
+        // Update variable order after aggregation
+        {
```
mmarx, line 83: Why is this inside a block?

nemo/src/execution/planning/aggregates.rs
```diff
@@ -0,0 +1,212 @@
...
+                variable_order_after_reordering.push(input_variable);
+            }
+
+            let reordering_column_indices: Vec<_> = variable_order_after_reordering
+                .as_ordered_list()
+                .iter()
+                .map(|variable| *variable_order.get(variable).unwrap())
+                .collect();
+
+            let reordering = ProjectReordering::from_vector(reordering_column_indices, 3);
```
matzemathics, line 50: Why is the domain size hardcoded to 3? Seems like a magic number to me.

nemo/src/execution/planning/aggregates.rs
mmarx: I see. Not sure if this really helps, for me just having the lines below one another without any blank lines is grouping enough, but we can leave it this way for all I care.

mmarx (approved):

matzemathics (approved):

PR #409, opened 2023-11-09
support wildcard terms

matzemathics:


mmarx:
Instead of getting an error about unsafe rules, we should have an explicit check and a more helpful error message for wildcards used in rule heads.

mmarx (changes requested):
As commented already, this still needs a helpful error message. Looks good otherwise!

mmarx (approved):

PR #411, opened 2023-11-09
simplify generic code in rdf triples reader

matzemathics:


nemo/src/io/formats/rdf_triples.rs
```diff
@@ -111,17 +111,13 @@ impl RDFTriplesReader {
...
 
-    fn read_with_buf_reader<'a, 'b, Reader, Parser, MakeParser>(
+    fn read_with_buf_reader<Parser>(
         &self,
-        physical_builder_proxies: &'b mut [PhysicalBuilderProxyEnum<'a>],
-        reader: &'b mut Reader,
-        make_parser: MakeParser,
+        physical_builder_proxies: &mut [PhysicalBuilderProxyEnum<'_>],
+        make_parser: impl FnOnce() -> Parser,
     ) -> Result<(), ReadingError>
```
mmarx, line 118: Since this no longer takes a `BufRead`, it should not be called `read_with_buf_reader` anymore, rather `read_with_triples_parser` or something like that?

mmarx (changes requested):

nemo/src/io/formats/rdf_triples.rs
matzemathics: that is a fair point.

mmarx (approved):

PR #412, opened 2023-11-09
cancel in progress ci runs, when new version is pushed

matzemathics:


.github/workflows/pr.yml
```diff
@@ -5,6 +5,10 @@ on:
     branches:
       - main
 
+concurrency:
+  group: ci-${{ github.head_ref }}
```
mmarx, line 9: It's unlikely that this will matter in practice, but I guess this should rather be `github.ref` instead, so that multiple PRs from the same source branch will not be in the same concurrency group.

mmarx (changes requested):

mmarx (approved):

Issue #417, opened 2023-11-17
Test fails randomly


aannleax:
Just had this error message:

```
---- io::formats::rdf_triples::test::example_1 stdout ----
thread 'io::formats::rdf_triples::test::example_1' panicked at nemo/src/io/formats/rdf_triples.rs:284:9:
assertion `left == right` failed
  left: "_:subje"
 right: "\"object1\""
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Only occurs once in like 20 runs.

monsterkrampe:
Trying to reproduce, I'm also getting another failed test sometimes:
```
---- io::formats::dsv::test::csv_with_ignored_and_faulty stdout ----
thread 'io::formats::dsv::test::csv_with_ignored_and_faulty' panicked at nemo-physical/src/dictionary/hash_map_dictionary.rs:90:22:
index out of bounds: the len is 5 but the index is 11688260230290059526
```

monsterkrampe:
The output in the original description also looks weird: `_:subje` looks very incomplete.
Seems to me like this could be related to the new dictionary (also because of the second failed test I found).

monsterkrampe:
Found another one:
```
---- dictionary::hash_map_dictionary::test::fetch_id stdout ----
thread 'dictionary::hash_map_dictionary::test::fetch_id' panicked at nemo-physical/src/dictionary/hash_map_dictionary.rs:394:9:
assertion `left == right` failed
  left: None
 right: Some(0)
stack backtrace:
   0: rust_begin_unwind
             at /rustc/ba7c7a301984967c8c13adb580ef9b86ba706a83/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/ba7c7a301984967c8c13adb580ef9b86ba706a83/library/core/src/panicking.rs:72:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
             at /rustc/ba7c7a301984967c8c13adb580ef9b86ba706a83/library/core/src/panicking.rs:279:5
   4: nemo_physical::dictionary::hash_map_dictionary::test::fetch_id
             at ./src/dictionary/hash_map_dictionary.rs:394:9
   5: nemo_physical::dictionary::hash_map_dictionary::test::fetch_id::{{closure}}
             at ./src/dictionary/hash_map_dictionary.rs:392:18
   6: core::ops::function::FnOnce::call_once
[...]

PR #422, opened 2023-11-21
Fix for issue #417

mkroetzsch:
Spurious concurrency errors are addressed by more aggressive locking. No test failures were detected across some 50-100 repetitions of the test cases.

If this is a performance problem, one could design a less congested data structure for StringBuffer (no vectors that can be moved entirely, esp. for page table; maybe fixed maximum of buffers with pre-allocated spaces that won't move either).


monsterkrampe:
Had the tests running in a loop for over an hour now and did not run into any failed tests so I think this is good to go :)

monsterkrampe (approved):

PR #429, opened 2023-11-28
parse nested map terms

matzemathics:


matzemathics:
@mmarx I am not sure where it is most appropriate, but somewhere we need to break the recursive chain of `impl FnMut(..) -> ..`, to make the type checker happy. I just went ahead and did this in the `parse_map_entry`, but I have no real opinion where it should be.

mmarx:
`parse_map_entry` is fine for now, I guess. I don't foresee much need to use this as a combinator.

mmarx (approved):

PR #430, opened 2023-11-28
Add ability to use arithmetic expressions and aggregates together

rlwww:


nemo/src/program_analysis/type_inference/type_requirement.rs
```diff
@@ -280,16 +302,35 @@ pub(super) fn requirements_from_aggregates_in_rules(
                 .iter()
                 .map(|term| {
                     if let PrimitiveTerm::Variable(Variable::Universal(identifier)) = term {
-                        if identifier.name().starts_with(AGGREGATE_VARIABLE_PREFIX) {
+                        // Add static output type to aggregate output variables and constructors using them
+
+                        let aggregate_output_variables_used = if identifier.name().starts_with(AGGREGATE_VARIABLE_PREFIX) {
+                            // If the term is a aggregate variable it self, add it
+                            vec![identifier]
+                        } else if identifier.name().starts_with(CONSTRUCT_VARIABLE_PREFIX) {
```
monsterkrampe, line 310: I'm wondering:
Should we maybe keep talking about variables instead of identifiers here and add methods to variables like:
`is_aggregate_variable()` and `is_constructor_variable()` (or similar)?
No one except the variables need to know that we distinguish them using certain prefixes and I would want to hide this as good as possible :)

nemo/src/program_analysis/type_inference/position_graph.rs
```diff
@@ -158,16 +176,19 @@ impl PositionGraph {
...
+                                    body_position.clone(),
+                                    head_position.clone(),
+                                    PositionGraphEdge::BodyToHeadSameVariable,
+                                );
+                            } else {
+                                debug_assert!(
+                                    matches!(body_variable, Variable::Universal(identifier) if identifier.0.starts_with(AGGREGATE_VARIABLE_PREFIX)),
+                                    "The iteration above went through all body atoms for non aggregate variables"
+                                )
+                            }
```
monsterkrampe, line 191: While working on #435 i realized that there should be probably an egde being added here as well to connect the aggregate output variable to the constructor (output) variable. In this sense, the aggregate output variable behaves like a body variable I think.

monsterkrampe (changes requested):
Could imagine that there is some need for discussion on what the type inference should do here. I'm also not quite sure in how far e.g. constructors and aggregates may depend on each other now so maybe let's discuss this offline.
Also, keep in mind the changes from #436 (which I hope to merge soon).

monsterkrampe (approved):
LGTM :)

PR #432, opened 2023-11-28
fix off-by-one-error in cut layer calculation

matzemathics:


mmarx (approved):

matzemathics:
@mmarx Why does github give me a "This branch cannot be rebased due to too many changes". Main is up to date, isn't it?

PR #455, opened 2024-03-13
Parser

### LINKED ISSUE #219: Redesign error messages
Rework the way error messages are displayed to the user. Errors should ideally be easy to identify, be concise, yet still understandable. In particular, we might want to simplify the formatting of parse errors to
* more clearly show the exact portion of the input that caused the error, and
* likely only show the top-level error, unless verbosity is increased.

### LINKED ISSUE #346: Public API for adding sources `Program`s
It would be nice if one could programmatically add sources to a `Program`, which may not be backed by a file, but by an arbitrary stream of tuples (maybe generated ones). Especially for the python integration this would be beneficial, eg.:
```python
program = load_string('result(?value) :- data("key", ?value).')
program.add_souces("data", [["something", 1], ["key", 42]])
```

### LINKED ISSUE #443: Problems when evaluating multiple levels of arithmetic definitions
The following program works:
```
@declare A(any, integer) .
@declare B(any, integer) .
@declare C(any, integer) .

C(?x, ?y) :- A(?x, ?z), B(?x, ?v), ?y = ?z + ?v, ?y > 10 .
```
However, when I try to replace `?y` in the head with `?y * 5`, I run into various problems.

Option 1
```
C(?x, ?y * 5) :- A(?x, ?z), B(?x, ?v), ?y = ?z + ?v, ?y > 10 .
```
throws
```
        Parse error on line 5, column 1: Complex term "?y * 5" uses an undefined variable "y".
```

Option 2
```
C(?x, ?u) :- A(?x, ?z), B(?x, ?v), ?y = ?z + ?v, ?y > 10, ?u = ?y * 5 .
```
throws
```
Parse error on line 5, column 1: The variable "u" must only depend on variables that occur in a positive body literal.
```

Option 3
```
C(?x, (?z + ?v) * 5) :- A(?x, ?z), B(?x, ?v), ?z + ?v > 10 .
```
works, but requires duplicating (or triplicating ...) terms, which is less intuitive, in particular if ?y has a clear meaning in the application.

I can live with Option 3 for now, but in the long run I would like to use more complex expressions for which both Option 1 (arithmetics over defined variables) and Option 2 (multiple levels of variable definitions) would be very useful.

### LINKED ISSUE #447: Error output: Add line number & relation
When receiving an error like:
`error: Overloading of predicate names by arity is currently not supported.`
it would be good, to output the line number and relation which caused he error.

### LINKED ISSUE #465: Built-in functions should work as atoms
Built-in functions should also work as atoms. In particular, boolean functions, such as `isIri`, should be allowed in rule bodies without having to compare their output to "true". For non-boolean functions, this can also be allowed by using the [*Effective Boolean Value* (EBV)](https://www.w3.org/TR/sparql11-query/#ebv) to decide if a value counts as "true", in essence:
- Strings are false when empty, otherwise true.
- Numbers are false when 0, otherwise true (we don't have NaN in Nemo).
- All other values, including IRIs and language-tagged strings, are always an error (which is effectively false when used in atoms, but should be interpreted differently if used in boolean `NOT`).

This issue does not ask for allowing these built-in atoms to be negated. If this were supported syntactically, it should behave as if the function had been wrapped in `NOT` (but also using the EBV to interpret the input to `NOT`, as, in general, all boolean functions should do, which would also be another issue but less pressing).

### LINKED ISSUE #466: Logical model should support nested structured terms with variables
The logical model currently supports arbitrary data values, which include nested structured ground terms (based on tuples and maps). This is also parsed, and actively used in `@import` and `@export` parameters.

The logical model needs to be extended to support similar terms even if not ground. Such terms can contain variables, and are therefore not captured in the datavalue model of nemo-physical. A possible economic solution would be to make the latter generic, so that it can also be instantiated with types that allow variables and ground datavalues to be used as fillers.

This is also a prerequisite for #464 and other issues that require higher-arity built-in functions.

### LINKED ISSUE #522: Fix reported positions by the parser of non-ascii characters
Our new error reporting relies on associating syntactic objects with their position in the text file, However, as of right now the position does not properly take into account non-ascii characters, which leads to wrong sections being highlighted in error messages.

### LINKED ISSUE #539: Wasm Code should use new error reporting
When running a program through wasm, the pretty error reports are not used yet. Instead, we just print debug messages for the errors.
This should be adjusted to use the new error reporting from the `parser` branch.

matzemathics:
Closes #466.
Closes #465.
Closes #447.
Closes #443.
Closes #346.
Closes #219. 
Closes #539.
Closes #522.


nemo/src/io/parser.rs
```diff
@@ -2431,60 +2443,288 @@ mod new {
...
-    macro_rules! tag_token (
-        ($func_name:ident, $tag: expr) => (
-            fn $func_name(tokens: Tokens) -> IResult<Tokens, Tokens> {
-                verify(take(1usize), |t: &Tokens| t.tok[0].kind == $tag)(tokens)
-            }
-        )
-    );
+    fn combine_spans<'a>(span1: Span<'a>, span2: Span<'a>) -> Result<Span<'a>, str_concat::Error> {
+        // SAFETY: The concatenation of strings is okay, because they originated from the same source string.
+        // The raw offset is okay, because the offset of another span is used.
```
mmarx, line 2448: pretty sure this is unsound, since the requirement that the two spans are from the same allocation is not actually enforced anywhere. As such, `combine_spans` should itself be `unsafe`.

nemo/src/io/lexer.rs
```diff
@@ -166,6 +222,7 @@ impl std::fmt::Display for TokenKind {
             TokenKind::PrefixIdent => write!(f, "PrefixIdent"),
             TokenKind::Illegal => write!(f, "Illegal"),
             TokenKind::Eof => write!(f, "Eof"),
+            TokenKind::Error => write!(f, "\x1b[1;31mError\x1b[0m"),
```
mmarx, line 225: Please don't use raw escape sequences, and also don't colour unconditionally. In `nemo-cli` we already use [`colored::Colorize`](https://docs.rs/colored/latest/colored/trait.Colorize.html) for this, which handles, e.g., the `NO_COLOUR` environment variable and terminal capabilities.

aannleax (approved):
Should be ready to merge now

PR #457, opened 2024-03-27
Bug/452

### LINKED ISSUE #452: negation error with constants inside the negated atom
Using constants in negated atoms does not work, when there is more than 1 negated atom per rule.
In the example I use string constants, but it is the same with integers.

1.  Error, when using one constant `"a"` in 2 negated atoms:
```
C(?a,?a0,?a1):-
    A(?a,?a0,?a1),
    ~ B(?a,?a0),
    ~ B("a",?a1).
```
2.   Error, when using two constants in 2 negated atoms:
```
C(?a,?a0,?a1):-
    A(?a,?a0,?a1),
    ~B("a",?a0),
    ~B(?"a",?a1).
```

The error message is:
```
thread 'main' panicked at nemo/src/execution/planning/arithmetic.rs:38:26:
Variable order must contain an entry for every variable.
```

The error also triggers, when the two negated atoms are referring to  different relations (i.e. replace the second negated atom with `~D("a",?a1)` and it will still not work).

aannleax:
Fixes #452.


nemo/src/execution/planning/operations/negation.rs
```diff
@@ -36,9 +37,18 @@ pub(crate) fn node_negation(
...
 
-            // We simply apply all constraints to this node
-            // Constraints which do not reference this atom will be filtered in the physical layer
-            let node_filtered = node_filter(plan, variable_translation, node, subtracted_filters);
+            // We only keep those constraints that can be evaluated within the current atom
+            let filters = subtracted_filters
+                .iter()
+                .filter(|filter| {
+                    filter.variables().all(|variable| {
+                        subtract_markers.contains(variable_translation.get(variable).unwrap())
```
monsterkrampe, line 45: Can we avoid the `unwrap()`?

monsterkrampe (changes requested):

aannleax:
Replaced the unwrap with an expect :)

monsterkrampe (approved):

Issue #448, opened 2023-12-20
Problem with new variables in body and inequality


Laschoking:
Variables which are created in the body cannot be used for comparison;  `<,>,!=` don't work on them.
This example fails, where `?b` is created in the body:
```
A(1).
B(?b):-
    A(?a),
    ?b = ?a +1,
    ?b > 2.
```
with 
`thread 'main' panicked at nemo/src/execution/planning/arithmetic.rs:38:26:
Variable order must contain an entry for every variable.`

The panic seems to happen for each new variable that is created
 i.e. if we replace ` ?b = ?a +1,` with `?b=2` the problem persists, but not for `?b = ?a`.

PR #458, opened 2024-03-28
Add error message for filtered derived variables

aannleax:
Adresses #448.

The way our planning works currently does not allow for rules such as 
```
A(1).
B(?b):-
    A(?a),
    ?b = ?a +1,
    ?b > 2.
```

This is because we first apply filters, and only later compute bindings for "fresh variables". It would be possible to then apply another filter step, but tbh. I feel that this should be solved in a more general way (a similar problem prevents us from having calculation inside aggregates and limits us to only "one layer" of derived variables). So for now I decided to simply throw an error in this case (instead of crashing). To not break existing testcases I also recognize every variable as safe that is equal (via `?x = ?y` filter) to a safe variable.


monsterkrampe (approved):
LGTM :)

PR #459, opened 2024-03-28
Bug/428

### LINKED ISSUE #428: triescan_minus potential bug
I just run the following program:
```
B(1).
B(-1).
Assert(1):-
    B(?a),
    ?a < 0.

A(?a):-
    B(?a),
    ~Assert(1).
```
which returned this error message:
```
thread 'main' panicked at nemo-physical/src/tabular/operations/triescan_minus.rs:94:22:
There should be at least one layer that is used.
```
Is there something I'm missing or is this a bug?

aannleax:
Fixes #428.

The bug concerns rule like `NonPositive(?r) :- Input(?r), ~Assert(positive).` and is related to the fact that negative atoms are projected to the variables of the positive atoms, resulting in a zero-column temporary table, which is then subtracted from the main join. 

As we now handle zero-column tables, I just needed to add this special case to the generator function of `TrieScanSubtract`, which now emits an empty result if a non-empty zero -column table is subtracted.


matzemathics (approved):
Looks good, maybe consider renaming `filter` to `constraint` though, due to the unfortunate clash with rust iterator terminology.

PR #470, opened 2024-04-10
Add binary builtin functions for numeric max/min

monsterkrampe:
addresses #442

PR #471, opened 2024-04-10
More builtins / Ternary and N-ary Builtins

### LINKED ISSUE #442: Min/max as built-in arithmetic functions
Do you plan to support "min" and "max" for arithmetic data types? Then one could write something like
```
E(?x, ?y) :- A(?x, ?z), B(?x, ?u), C(?x, ?v), D(?x, ?w), ?y = min(?z,?u,?v,?w) .
```
I could also write this as 4 different rules with "<="-constraints, but that would not be very readable.

### LINKED ISSUE #463: DATAYPE function should return IRI, not string
The built-in function DATATYPE returns the datatype IRI as a string. This should be an IRI instead. This is also how it is specified in SPARQL: https://www.w3.org/TR/sparql11-query/#func-datatype

### LINKED ISSUE #464: SUBSTR function should support optional third length parameter
The SUBSTR function currently just supports two parameters: the string and the starting index. This form is also supported in SPARQL, but there should be an additional form that takes a third parameter for the length, so that arbitrary can be returned.

aannleax:
Implements ternary and n-ary builtin functions and generalizes some existing binary functions to nary functions. For example:

```
result(concat, ?R) :- strings(?A, ?B), ?R = CONCAT(?A, " ", ?B).
result(subStringLength, ?R) :- strings(?A, _), ?R = SUBSTRING(?A, 1, 3).

result(sum, ?R) :- doubles(?A, ?B, ?C), ?R = SUM(?A, ?B, ?C).
result(prod, ?R) :- doubles(?A, ?B, ?C), ?R = PROD(?A, ?B, ?C).
result(min, ?R) :- doubles(?A, ?B, ?C), ?R = MIN(?A, ?B, ?C).
result(max, ?R) :- doubles(?A, ?B, ?C), ?R = MAX(?A, ?B, ?C).
result(luka, ?R) :- numbers(?A, ?B, ?C), ?R = LUKA(?A, ?B, ?C) .
result(bitand, ?R) :- integers(?A, ?B, ?C), ?R = BITAND(?A, ?B, ?C).
result(bitor, ?R) :- integers(?A, ?B, ?C), ?R = BITOR(?A, ?B, ?C).
result(bitxor, ?R) :- integers(?A, ?B, ?C), ?R = BITXOR(?A, ?B, ?C).
```

Closes #470.
Closes #464.
Closes #463.
Closes #442.


nemo-physical/src/function/definitions/boolean.rs
```diff
@@ -5,40 +5,36 @@ use crate::{
...
-        bool_pair_from_any(parameter_first, parameter_second)
-            .map(|(first_bool, second_bool)| AnyDataValue::new_boolean(first_bool && second_bool))
+impl NaryFunction for BooleanConjunction {
+    fn evaluate(&self, parameters: &[AnyDataValue]) -> Option<AnyDataValue> {
+        if let Some(bools) = bool_list_from_any(parameters) {
+            let result = bools.into_iter().all(|bool| bool);
+            Some(AnyDataValue::new_boolean(result))
+        } else {
+            None
+        }
```
monsterkrampe, line 37: This could still use `Option::map`.

nemo-physical/src/function/definitions/numeric.rs
```diff
@@ -242,6 +296,83 @@ impl BinaryFunction for NumericDivision {
...
+impl NaryFunction for BitAnd {
+    fn evaluate(&self, parameters: &[AnyDataValue]) -> Option<AnyDataValue> {
+        if let Some(list) = NumericList::from_any_list(parameters) {
+            match list {
+                NumericList::Integer(integers) => numeric_bitwise_and(&integers),
+                _ => None,
+            }
+        } else {
+            None
+        }
```
monsterkrampe, line 316: Similar to above, this could use `Option::and_then`.
This actually applies to most (or even all) of the added/changed functions below.

nemo-physical/src/function/tree.rs
```diff
@@ -740,6 +816,111 @@ where
...
+    }
+
+    // NaryOperation::BitAnd => todo!(),
+    //             NaryOperation::BitOr => todo!(),
+    //             NaryOperation::BitXor => todo!(),
+    //             NaryOperation::NumericSum => todo!(),
+    //             NaryOperation::NumericProduct => todo!(),
+    //             NaryOperation::NumericMinimum => todo!(),
+    //             NaryOperation::NumericMaximum => todo!(),
+    //             NaryOperation::NumericLukasiewicz => todo!(),
```
monsterkrampe, line 842: Can this be removed?

monsterkrampe (approved):

PR #473, opened 2024-04-24
Builtin typecast

aannleax:
Automatically cast numeric builtins to to double if needed

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,4 +1,4 @@
-[build] 
+[features] 
 # See https://rustwasm.github.io/docs/wasm-bindgen/web-sys/unstable-apis.html
 # Required for https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.FileSystemSyncAccessHandle.html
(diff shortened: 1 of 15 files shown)


nemo/src/execution/planning/plan_head_restricted.rs
```diff
@@ -113,6 +113,7 @@ impl HeadStrategy for RestrictedChaseStrategy {
         current_plan: &mut SubtableExecutionPlan,
         variable_translation: &VariableTranslation,
         node_matches: ExecutionNodeRef,
+        _node_aggregates: Option<ExecutionNodeRef>,
```
monsterkrampe, line 116: Why is this here? The underscore seems to indicate that this is not used(?)

monsterkrampe (approved):
Just one question about a line that might be obsolete.

nemo/src/execution/planning/plan_head_restricted.rs
aannleax: Aggregation unterstützen wir bislang nicht zusammen mit existentiellen Regeln

PR #479, opened 2024-05-22
Fix petgraph trace

monsterkrampe:
Fixes https://github.com/knowsys/nemo-web/issues/5


nemo/src/execution/tracing/trace.rs
```diff
@@ -207,10 +207,10 @@ impl TraceTreeRuleApplication {
     }
 
     /// Get the [`ChaseFact`] that was produced by this rule application.
```
mmarx, line 209: Documentation needs an update ;)

mmarx (changes requested):

mmarx (approved):

PR #485, opened 2024-06-04
Reset type when going into a new layer

aannleax:
I think the previous fix (#482) to the mixed type problem was not correct and would lead to unneeded iterations in a cut layer.


matzemathics (approved):
Ok, this now makes more sense.

PR #486, opened 2024-06-04
Handle HttpResources as raw bytes, not as text

mmarx:
Trying to handle these as text messes with, e.g., the GZip header, leading to decompression failures.


matzemathics (approved):
good catch!

PR #487, opened 2024-06-04
Fix detection of special case for constant evaluation

aannleax:
We have special handling for expressions like `?x = 5`, which uses the `seek`-function to quickly skip over values. However, the detection of this case was previously broken, so that we instead evaluated a `= 5` function for each entry.

This PR should fix this problem and also has the observed performance benefits of #484.

### DIFF
--- a/nemo-physical/src/tabular/operations/filter.rs
+++ b/nemo-physical/src/tabular/operations/filter.rs
@@ -131,11 +131,7 @@ impl GeneratorFilter {
     /// Helper function that takes a list of boolean [Filter]s
     /// and constructs a [Filter] that represent its conjuction.
     fn fold_filters(filters: Vec<&Filter>) -> Filter {
-        if filters.len() == 1 {
-            filters[0].clone()
-        } else {
-            Filter::boolean_conjunction(filters.into_iter().cloned().collect::<Vec<_>>())
-        }
+        Filter::boolean_conjunction(filters.into_iter().cloned().collect::<Vec<_>>())
     }
 
     /// Compute the [FilterAssignment] from a list of [Filters].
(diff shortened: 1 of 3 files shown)


nemo-physical/src/tabular/operations/filter.rs
```diff
@@ -131,7 +131,11 @@ impl GeneratorFilter {
     /// Helper function that takes a list of boolean [Filter]s
     /// and constructs a [Filter] that represent its conjuction.
     fn fold_filters(filters: Vec<&Filter>) -> Filter {
-        Filter::boolean_conjunction(filters.into_iter().cloned().collect::<Vec<_>>())
+        if filters.len() == 1 {
+            filters[0].clone()
+        } else {
+            Filter::boolean_conjunction(filters.into_iter().cloned().collect::<Vec<_>>())
```
monsterkrampe, line 137: Would it make sense to handle this special case within the `Filter::boolean_conjunction` function?

monsterkrampe (changes requested):
See comment above.

nemo-physical/src/tabular/operations/filter.rs
aannleax: I would like to keep the constructors simple. Doing such "optimizations" for conjunctions would probably also mean one would have to do them for all functions (for consistency :)). 

Also it should be fine for most cases to have a conjunction of one thing; it just breaks something in this specific instance, so I think it is appropriate to handle this here

aannleax: I have now included that check for all n-ary functions

monsterkrampe (approved):
Nice, thanks!
I'm wondering if there should also be a special case for `parameters.len() == 0` but maybe we do not need to be concerned about this case. The current code should certainly cover this to a satisfying extent.

mmarx:
> Nice, thanks! I'm wondering if there should also be a special case for `parameters.len() == 0` but maybe we do not need to be concerned about this case.

Once we also handle that, I think it'll make sense to pull this functionality out into some generic piece of code called by the constructors – for now, the current approach is fine.

PR #491, opened 2024-06-06
Add unary builtin function for reversing a string

aidan-bailey:


aannleax (approved):
Looks good :)

While looking at these builts-ins again, I noticed that all of our string functions do not support non-ascii characters (including this new operation). But I think this should be uniformly handled as part this new issue: #493.

Issue #493, opened 2024-06-10
String methods should handle non-ascii characters


aannleax:
The current implementation of the string functions does not properly handle non-ascii characters. For example
```
word("loẅks").
count(STRLEN(?X)) :- word(?X).
```

returns `count(7)`.

Possibly, this is a good crate to use: https://docs.rs/unicode-segmentation/latest/unicode_segmentation/

aidan-bailey:
I can have a look at this?

aannleax:
> I can have a look at this?

Sure, go ahead :)

PR #495, opened 2024-06-11
Add Unicode support to String functions

aidan-bailey:
Addresses #493 

This turned out to be a lot more involved than expected.
I could not find a plug-and-play replacement for unicode strings, so as suggested I used `unicode_segmentation` (quite a lot...)

What is a *grapheme*? "the smallest meaningful contrastive unit in a writing system." - Oxford Languages (Google)

In ASCII, all graphemes are simple atomic characters.

In Unicode, graphemes can be atomic characters like in ASCII, but they can also consist of multiple characters allowing for more complex representations.

For example, the unicode grapheme "ẅ" actually consists of the symbols "w" and "\u{308}", i.e. "w\u{308}".

Thus, if we use the conventional string method `"ẅ".len()` , it evaluates to `2`, not `1` as one would probably expect.

`unicode_segmentation` allows us to construct a vector of unicode graphemes of any string. We can then compute the length of the grapheme vector of "ẅ" and get the result we expect, `1`.

This problems extends to other methods. For example, both `"ẅ".contains("w")` and `"ẅ".begins_with("w")` evaluate to `true`. Thus, all methods that involve any kind of indexing were fixed by using unicode grapheme vectors (and I added a bunch of unit tests for them too!)

Methods Affected:
- `StringLength` 
- `StringStarts` 
- `StringEnds`
- `StringLowercase`
- `StringUppercase`
- `StringBefore`
- `StringAfter`
- `StringContains`

Since I've had to essentially reimplement a bunch of string methods, there is a lot of new indexing shenanigans (particularly in `unicode_find`.) Ideally a vector would have `.contains` and `.find` methods which would take another vector as an argument and function similarly to the string methods of the same name but alas. I have tried to write this method with an appropriate readability-to-efficiency balance.

If anyone has suggestions of better/more appropriate ways to do this, please do suggest them! Of course we can also optimise this in the future.


nemo-physical/src/function/definitions/string.rs
```diff
@@ -11,6 +12,57 @@ use super::{
     BinaryFunction, FunctionTypePropagation, NaryFunction, TernaryFunction, UnaryFunction,
 };
 
+/// Unicode friendly version of [String] find
+///
+/// Returns the index of the first occurrence of the second parameter in the first parameter.
+/// or `None` if the second parameter is not found.
+fn unicode_find(parameter_first: &String, parameter_second: &String) -> Option<usize> {
```
mmarx, line 19: These should really not be called `parameter_first` and `parameter_second`, but something descriptive, e.g., `needle` and `haystack` (or maybe `pattern` and `string`; in any case the name should already make clear which one is which – the order is already apparent from the signature anyways.)

aidan-bailey: I completely agree. I named the parameters that to be more consistent with the other method headers. Can it break the convention because it takes in `String`s instead of `AnyDataValue`s? Perhaps because it is essentially a utility method?

nemo-physical/src/function/definitions/string.rs
```diff
@@ -11,6 +12,17 @@ use super::{
...
 };
 
+/// Unicode friendly version of [String] find
+///
+/// Returns the index of the first occurrence of the needle in the haystack
+/// or `None` if the needle is not found.
+fn unicode_find(haystack: &String, needle: &String) -> Option<usize> {
+    let haystack_graphemes = haystack.graphemes(true).collect::<Vec<&str>>();
+    let needle_graphemes = needle.graphemes(true).collect::<Vec<&str>>();
+    (0..(haystack_graphemes.len() - needle_graphemes.len() + 1))
```
mmarx, line 22: The subtraction can underflow if `needle` is longer than `haystack`. Better use `saturating_sub` here.

mmarx (changes requested):
Spotted another potential bug, but otherwise looks good :)

nemo-physical/src/function/definitions/string.rs
aidan-bailey: Very nice catch! I think the solution is just doing a length check within the function.

If `needle` is longer than `haystack` and we do a `saturating_sub` (neat I did not know this function existed), we'll immediately evaluate `haystack_graphemes[i..i + needle_graphemes.len()]` which will result in an error since `needle` is longer than `haystack`.

PR #497, opened 2024-06-12
Refactor loop-based unicode grapheme comparison to be slice-based

aidan-bailey:
I apologise for this pedantic PR. The fact I refactored `unicode_find` to use slice-based comparison but left the loop-based comparison in `StringStarts` and `StringEnds` was really bugging me...


mmarx (approved):

mmarx:
> I apologise for this pedantic PR.

No need to apologise, this is good code! :)

PR #498, opened 2024-06-27
Add REGEX string builtin function

aidan-bailey:
Mirrors SPARQL ` xsd:boolean  REGEX (string literal text, simple literal pattern)` function.

I think a separate function should be created for the ternary variation
` xsd:boolean  REGEX (string literal text, simple literal pattern, simple literal flags)`
as was done for `SUBSTR` and `SUBSTRING`.

### DIFF
--- a/nemo-physical/src/function/definitions.rs
+++ b/nemo-physical/src/function/definitions.rs
@@ -34,7 +34,7 @@ use self::{
     },
     string::{
         StringAfter, StringBefore, StringCompare, StringConcatenation, StringContains, StringEnds,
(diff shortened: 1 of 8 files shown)


aidan-bailey:
@aannleax informed me of the potential performance hit we'll take from compiling the same regular expression multiple times. I've found a well-used library that addresses this: [regex-cache](https://crates.io/crates/regex-cache). Shall I integrate it or should we leave this optimising for future work?

aidan-bailey:
Hey hey hey, it's not abandoned, it's finished 😉

aannleax (approved):
Look good :)

PR #501, opened 2024-07-04
Bug/500 constant filters

### LINKED ISSUE #500: Exception (unreachable code) when forgetting '?' in rule body
Assume a small example: 
```
eq(?a) :- pair(?a,?b), ?a = ?b.
```

This runs and performs as intended, by introducing two typing mistakes into the rule, we will get an exception instead of an error message:
```
eq(?a) :- pair(?a,?b), a = b.
```
[error_msg.txt](https://github.com/user-attachments/files/16099878/error_msg.txt)

The current version is 0.5.2-dev (ca5ddd6aa6d089f71f4ad28a4a8285d6f94ba674), using rustc v1.80.0-nightly (see corresponding flake file)

aannleax:
I had code that computed the column on which a filter is going to be applied on. This depended upon the input variables, e.g. when computing `a(x, y, z), 2 * x > z`, the less-than filter should be applied on column `z` as the value for `x` has to be bound when computing the filter. The code that finds the last occurrence of a variable panicked if the filter did not contain any variable.

To fix this I simply assign constant filters to the first column.

Fixes #500.

### DIFF
--- a/nemo-physical/src/tabular/operations/filter.rs
+++ b/nemo-physical/src/tabular/operations/filter.rs
@@ -144,7 +144,7 @@ impl GeneratorFilter {
         let mut grouped_filters = HashMap::<OperationColumnMarker, Vec<&Filter>>::new();
 
         for filter in filters {
-            let marker = Self::find_last_reference(input, filter).unwrap_or(input[0].clone());
+            let marker = Self::find_last_reference(input, filter).unwrap_or(input[0]);
             grouped_filters.entry(marker).or_default().push(filter);
         }
 
--- a/nemo-physical/src/tabular/operations/filter.rs
+++ b/nemo-physical/src/tabular/operations/filter.rs
@@ -117,16 +117,20 @@ impl GeneratorFilter {
 
     /// Helper function that finds the [OperationColumnMarker] used in the given [Filter]
     /// that appears closest to the end of the given [OperationTable].
-    fn find_last_reference(table: &OperationTable, filter: &Filter) -> OperationColumnMarker {
+    ///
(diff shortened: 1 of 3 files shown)


nemo-physical/src/tabular/operations/filter.rs
```diff
@@ -125,7 +125,8 @@ impl GeneratorFilter {
             }
         }
 
-        unreachable!("Filter must only use markers from the table.")
+        // Compute constant columns in the first column
+        table[0]
```
monsterkrampe, line 129: From my high level understanding: Would it not be more intuitive to make this function return an `Option`?
Because in an example where the filter is only using constants, the column marker should really be `none`, no? (I could be wrong here, since I'm not totally sure what the `OperationColumnMarker` is.)

aannleax: I guess the way the function is named, it makes sense. I changed it accordingly and set the default value after the call.

aannleax:
Does anybody know a fix for the python binding?

The suggestion `consider adding a `#![type_length_limit="31589662"]` attribute to your crate` seems odd :D

mmarx:
> The suggestion `consider adding a #![type_length_limit="31589662"] attribute to your crate` seems odd :D

Maybe that part of the parser could be rewritten to produce a smaller closure, but for now, I'd just go with increasing the limit. This is only coming up now because the check for this in `rustc` was broken for quite a while: https://github.com/rust-lang/rust/pull/125507

matzemathics (approved):
lgtm

PR #503, opened 2024-07-10
Support Trace as DAG in GraphML format

monsterkrampe:


aannleax (approved):
No complaints from me :)

PR #525, opened 2024-09-04
Allow % (percent sign) in bare names

### LINKED ISSUE #454: Fix serialisation of IRIs with percent encoding
Currently iris like `<hello%20world>` are serialized as `hello%20world` in csv, which  is not interpreted as an iri upon reading.

monsterkrampe:
closes #454


aannleax (approved):
Looks good

PR #527, opened 2024-09-06
Compute and report dictionary size

### LINKED ISSUE #339: Calculate Memory Footprint of the Dictionary
Since the dictionary can consume quite a lot of memory it would be nice if it was visible in the output introduced by #338.

This would involve implementing the `ByteSized` trait for the dictionary and adding a new subblock to the `MemoryUsage` object created in the `memory_usage` function of the `TableManager`.

mkroetzsch:
This pull request fixes #339 and integrates the result in the reporting on the command line.


mkroetzsch:
My local branch had some old merges still going on. Sorry for the mess. The new parts are just my commits. This also removes ByteSize from nemo-physical and nemo, as discussed.

nemo-physical/src/management/database/storage.rs
```diff
@@ -114,13 +112,13 @@ impl TableStorage {
...
+    fn size_bytes(&self) -> u64 {
         match self {
             TableStorage::InMemory(trie) => trie.size_bytes(),
-            TableStorage::FromSources(sources) => {
-                sum_bytes(sources.iter().map(|source| source.size_bytes()))
-            }
-            TableStorage::Empty => ByteSize::b(0),
+            TableStorage::FromSources(sources) => sources
+                .iter()
+                .fold(0, |acc, source| acc + source.size_bytes()),
```
mmarx, line 120: why not something like this, which I think is more readable?

```rust
TableStorage::FromSources(sources) => sources.iter().map(
  |source| source.size_bytes()
).sum()
```

mkroetzsch: If you think it is worth your time, feel free to replace all uses of fold for summation accordingly. I won't make changes in this branch now.

monsterkrampe (approved):
LGTM :)

PR #531, opened 2024-09-12
Adjust language server and wasm for new parser

### LINKED ISSUE #511: Integrate new parser into Nemo Language Server
Implement the `LSPComponent` trait for each `ProgramAST` implementation
EDIT: Also replace usage of `ASTNode` in the language server by `ProgramAST`.

monsterkrampe:
I did changes in the language server that made sense to me but I cannot test this right now since some crucial parser methods are still unimplemented and therefor trying to parse (any) program panics.
Also some methods originally used in the wasm code do not exist anymore in the nemo crate. These would be necessary for a full adjustment but for only testing the language server, we are fine without them.

The `LSPComponent` trait is implemented generically for everything that has the `ProgramAST` trait. I hope that this makes sense. Since the language server relies on the `children` method, which only returns `Vec<&dyn ProgramAST>`, it is hard (pretty much impossible at the moment) to work with individual implementations for the `LSPComponent` trait anyway.

Feel free to already provide me with feedback if this looks reasonable.

Once finished, this will close #511.


monsterkrampe:
Note that I had to change some visibilities. Maybe these APIs should be reconsidered.

aannleax (approved):
Looks good. I like that the implementation of the LSPComponent trait is handled neatly in one file.

Issue #529, opened 2024-09-11
More efficient dictionaries for data that has a "frequent" part


mkroetzsch:
Many strings in our dictionaries consist of a "frequent" and a "rare" part, such as, datatypeIRI+literal, languageTag+string, prefix+localName. There should be dictionary implementations that exploit this to save memory by storing each frequent part only once instead of storing it in every entry that needs it. The mechanism can also be used for "typed" entries, where every "string" belongs to one of a small but possibly runtime-dependent number of available types (e.g., signatures of functions).

PR #532, opened 2024-09-13
More efficient dictionaries for data pairs

mkroetzsch:
This PR implements #529 by introducing a ranked pair dictionary that stores pairs using two-level mapping. The result is used in new datavalue (DV) dictionary implementations for IRIs (prefix/local), language tagged string (language tag/plain string), and other data literals (datatype IRI/literal value).

Based on local benchmarks, the new approach achieves significant memory savings for IRIs and datatype literals, but not for language-tagged strings (where the short tags do not justify the overhead), so only IRIs and datatypes use the ranked pair dictionary implementation in the meta dv dict, whereas langtagged strings use the previous string-based dv dict. Memory consumption on 100M non-unique Wikidata terms drops from 700M to 550M, for 1000M terms it drops from 5.6G to 4.3G. For more reasoning-heavy loads with much inferred data, the impact is smaller but still positive (e.g., SNOMED normalisation saves about 30M to 40M, with overall memory use around 400M including reasoning).

The PR also removes old pair implementations that did not accomplish an improvement for any of these cases due to symmetric handling of both parts in pairs. Moreover, other old dictionaries (pre Nemo 0.5) have been moved to a sub-module, kept only for benchmarking.


nemo-physical/src/dictionary/ranked_pair_dictionary.rs
```diff
@@ -0,0 +1,499 @@
...
+/// very small cache sizes, where the most efficient way of getting an item is
+/// to iterate over all entries.
+#[derive(Debug)]
+struct LruArray {
+    cache: Vec<(Vec<u8>, usize)>,
+    top: usize,
+    len: usize,
+}
+impl LruArray {
+    fn create(capacity: usize) -> Self {
```
mmarx, line 48: The constructor should be called `new`, cf. the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines/predictability.html). Also, `capacity` should be `std::num::NonZero<usize>` to make this sound.

nemo-physical/src/dictionary/ranked_pair_dictionary.rs
```diff
@@ -0,0 +1,499 @@
...
+    /// Looks for a given pair and returns `Some(id)` if it is in the dictionary,
+    /// and `None` otherwise. The special value [super::KNOWN_ID_MARK] will be returned
+    /// if the pair was marked but not actually inserted.
+    pub(crate) fn pair_to_id(&self, frequent: &[u8], rare: &[u8]) -> Option<usize> {
+        if let Some(fid) = self.frequent_dict.bytes_to_id(frequent) {
+            self.pair_dict.bytes_to_id(&Self::pair_bytes(fid, rare))
+        } else {
+            None
+        }
+    }
```
mmarx, line 207:
```rust
        let fid = self.frequent_dict.bytes_to_id(frequent);
        self.pair_dict.bytes_to_id(&Self::pair_bytes(fid?, rare))
```

nemo-physical/src/dictionary/ranked_pair_dictionary.rs
```diff
@@ -0,0 +1,499 @@
...
+            }
+
+            rare = vec![0; pair_bytes.len() - fid_len];
+            rare.copy_from_slice(&pair_bytes[fid_len..]);
+            let frequent = self.frequent_dict.id_to_bytes(fid).expect("");
+
+            Some((frequent, rare))
+        } else {
+            None
+        }
```
mmarx, line 254:
```rust
        let pair_bytes = self.pair_dict.id_to_bytes(id)?;
        let mut rare: Vec<u8>;
        let fid: usize;
        let fid_len: usize;
        match pair_bytes[0] & FTYPE_BIT_MASK {
            0b0000_0000 => {
                // 1 byte for frequent id
                fid = pair_bytes[0] as usize;
                fid_len = 1;
            }
            0b0100_0000 => {
                let mut fid_bytes: [u8; USIZE_BYTES] = usize::to_be_bytes(0);
                fid_bytes[USIZE_BYTES - 1] = pair_bytes[1];
                fid_bytes[USIZE_BYTES - 2] = pair_bytes[0] & FID_BIT_MASK;
                fid = usize::from_be_bytes(fid_bytes);
                fid_len = 2;
            }
            0b1000_0000 => {
                let mut fid_bytes: [u8; USIZE_BYTES] = usize::to_be_bytes(0);
                fid_bytes[USIZE_BYTES - 3..].copy_from_slice(&pair_bytes[0..=2]);
                fid_bytes[USIZE_BYTES - 3] &= FID_BIT_MASK;
                fid = usize::from_be_bytes(fid_bytes);
                fid_len = 3;
            }
            0b1100_0000 => {
                let mut fid_bytes: [u8; USIZE_BYTES] = usize::to_be_bytes(0);
                fid_bytes[USIZE_BYTES - 4..].copy_from_slice(&pair_bytes[0..=3]);
                fid_bytes[USIZE_BYTES - 4] &= FID_BIT_MASK;
                fid = usize::from_be_bytes(fid_bytes);
                fid_len = 4;
            }
            _ => unreachable!("The bit mask ensures that values match one of the other arms"),
        }

        rare = vec![0; pair_bytes.len() - fid_len];
        rare.copy_from_slice(&pair_bytes[fid_len..]);
        let frequent = self.frequent_dict.id_to_bytes(fid).expect("");

        Some((frequent, rare))
```

nemo-physical/src/dictionary/ranked_pair_dv_dict.rs
```diff
@@ -0,0 +1,133 @@
...
+        panic!("string-pair dictionaries cannot make fresh nulls");
+    }
+
+    fn datavalue_to_id(&self, dv: &AnyDataValue) -> Option<usize> {
+        if let Some((frequent, rare)) = C::dict_pair(dv) {
+            self.string_pair_dict
+                .pair_to_id(frequent.as_str(), rare.as_str())
+        } else {
+            None
+        }
```
mmarx, line 74:
```rust
        let (frequent, rare) = C::dict_pair(dv)?;
        self.string_pair_dict
            .pair_to_id(frequent.as_str(), rare.as_str())
```

nemo-physical/src/dictionary/ranked_pair_dv_dict.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    fn id_to_datavalue(&self, id: usize) -> Option<AnyDataValue> {
+        if let Some((frequent, rare)) = self.string_pair_dict.id_to_pair(id) {
+            C::pair_to_datavalue(frequent.as_str(), rare.as_str())
+        } else {
+            None
+        }
```
mmarx, line 82:
```rust
        let (frequent, rare) = self.string_pair_dict.id_to_pair(id)?;
        C::pair_to_datavalue(frequent.as_str(), rare.as_str())
 ```

File: nemo-physical/src/dictionary/string_dv_dict.rs [hunk 6]
```diff
@@ -5,26 +5,109 @@
...
+    }
+    let string1: String;
+    let string2: String;
+    let marker = string.as_bytes()[0] as usize;
+    if marker > 0 {
+        if let Some(iri_string) = string.get(1..marker + 1) {
+            string2 = iri_string.into();
+        } else {
+            return None;
+        }
```
mmarx, line 69:
```rust
        let iri_string = string.get(1..marker + 1)?;
        string2 = iri_string.into();
```

nemo-physical/src/dictionary/meta_dv_dict.rs
```diff
@@ -305,20 +305,23 @@ impl MetaDvDictionary {
                     return;
                 }
                 dict = Box::new(LangStringDvDictionary::new());
+                //dict = Box::new(LangStringSplittingDvDictionary::new());
```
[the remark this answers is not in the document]
mkroetzsch: Well, it is the place where one can drop in the old dict if desired for testing. Since the code is still there (also commented out in the module file), I thought it makes sense to handle the actual use similarly.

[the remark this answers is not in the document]
mmarx: Swapping the dictionaries also requires changing the imports, so it's not that this would be a single-line change, but fair enough if you absolutely want to keep it.

mmarx (approved):

nemo-physical/src/dictionary/string_dv_dict.rs
```diff
@@ -120,6 +203,12 @@ impl<C: DvConverter> DvDict for StringBasedDvDictionary<C> {
 
 impl<C: DvConverter> ByteSized for StringBasedDvDictionary<C> {
     fn size_bytes(&self) -> u64 {
+        // DEBUG
+        // println!(
+        //     "String DV Dict for {:?}: size {}",
+        //     C::supported_value_domain(),
+        //     self.string_dict.size_bytes()
+        // );
```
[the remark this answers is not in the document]
mkroetzsch: Yeah, not sure about this. I think some more work is needed to make these pieces good enough for becoming log messages (also in the order of computation and output). Performance is not an issue in this place, but it is not clear if summarising logs should be printed as part of a method that returns memory usage. If so, then we should probably do this in all uses of this method, but it feels wrong.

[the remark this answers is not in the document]
mmarx: We could also use log messages with a particular log target, then one could toggle these on or off systematically at runtime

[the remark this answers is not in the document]
mkroetzsch: Ok, I did clean it up after all and use log now ...

PR #536, opened 2024-09-18
LSP: Better DocumentSymbols; Syntax Highlighting

### LINKED ISSUE #533: Check correct Generation of DocumentSymbols in Language Server for new Parser
As a follow up to #511, we should double check that the generated `DocumentSymbol`s are sensible and behave as expected in VSCode.

monsterkrampe:
closes #533


aannleax (approved):
Loos good :)

PR #537, opened 2024-09-18
Ditch `nom_locate`

matzemathics:
This implementation of `Span` uses a pointer to the start of the allocation instead of an implied pointer through a `usize` offset, to make provenance/aliasing tracking work. This means that computing the offset goes from being a lookup to being a subtraction, the performance implications of which I have not tested. Other layouts are possible, but either use more memory, ore make computing the fragment more costly.

@aannleax is there an existing benchmark to test the performance implications? I don't expect them to be severe, but that's just me guessing.


mmarx (approved):
LGTM. I haven't tested the performance implications, but parsing should generally not be the bottleneck, as programs are usually much smaller than data.

PR #541, opened 2024-10-02
sync wasm-bindgen version between cargo and nix

matzemathics:
`nix flake check` should now work again.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -25,11 +25,7 @@ rec {
       channels.nixpkgs.overlaysBuilder = channels: [
         rust-overlay.overlays.default
         (final: prev: {
-          wasm-bindgen-cli = channels.nixpkgs-unstable.wasm-bindgen-cli.override {
-            version = "0.2.93";
-            hash = "sha256-DDdu5mM3gneraM85pAepBXWn3TMofarVR4NbjMdz3r0=";
-            cargoHash = "sha256-birrg+XABBHHKJxfTKAMSlmTVYLmnmqMDfRnmG6g/YQ=";
-          };
+          inherit (channels.nixpkgs-unstable) wasm-bindgen-cli;
         })
       ];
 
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -27,9 +27,9 @@ rec {
         rust-overlay.overlays.default
         (final: prev: {
           wasm-bindgen-cli = channels.nixpkgs-unstable.wasm-bindgen-cli.override {
-            version = "0.2.92";
-            hash = "sha256-1VwY8vQy7soKEgbki4LD+v259751kKxSxmo/gqE6yV0=";
-            cargoHash = "sha256-aACJ+lYNEU8FFBs158G1/JG8sc6Rq080PeKCMnwdpH0=";
+            version = "0.2.93";
+            hash = "sha256-DDdu5mM3gneraM85pAepBXWn3TMofarVR4NbjMdz3r0=";
+            cargoHash = "sha256-birrg+XABBHHKJxfTKAMSlmTVYLmnmqMDfRnmG6g/YQ=";
           };
```
mmarx, line 33: unstable already has `wasm-bindgen-cli` at 0.2.93, so no need to override here, simply do 
```nix
inherit (channels.nixpkgs-unstable) wasm-bindgen-cli;
```

mmarx (changes requested):

flake.nix
matzemathics: However that solution might then break again in the future, right?

mmarx: Yeah, that will break once unstable upgrades `wasm-bindgen-cli`, but then we can just bump the bound on the cargo side to fix this.

matzemathics: So for some reason this version fails on my machine with the same error we got in the ci before, which suggests it is using the wrong version. Do I need to update something?

mmarx: You'll likely need to run `nix flake update --commit-lock-file`

matzemathics: `API rate limit exceeded for ***.***.***.***` :sweat_smile:

mmarx: Haha, last time I saw that, I was traewelling with DB :D

I've pushed the updated lock, could you please check that this now works for you?

mmarx (approved):

flake.nix
matzemathics: Seems to work now.

PR #542, opened 2024-10-02
Release 0.6

aannleax:


CITATION.cff
```diff
@@ -56,7 +56,7 @@ keywords:
   - Existential Rules
 license: Apache-2.0
 commit: 6ccd3fcc12d57fbac5e3d7e5c5547e72a610d7f7
-version: 0.5.1
+version: 0.6.0
```
mmarx, line 59: If you bump the version, you also need to bump the commit, 6ccd3fcc12d57fbac5e3d7e5c5547e72a610d7f7 corresponds to the v0.5.1 tag.

mmarx (changes requested):
Maybe also run `cargo update`, to make sure we're building against current dependencies?

CITATION.cff
aannleax: Good point

mmarx (approved):

PR #545, opened 2024-10-29
Remove '-' as valid characters for 'names'

### LINKED ISSUE #543: Variable names get confused with arithmetic operations
The following rule produces an unsafe-variable error:

```
r(?X-1) :- a(?X).
```

This is because X-1 is a valid variable name, so the parser does not recognize this as an subtraction.

aannleax:
Removed `-` as a valid characters that can appear in "names" like variable or prefix names. This lead to confusion with the infix operator `-` for subtraction. 

Closes #543


monsterkrampe (approved):

nemo/src/parser/ast/token.rs
```diff
@@ -316,7 +316,7 @@ impl<'a> Token<'a> {
             ParserContext::token(TokenKind::Name),
             recognize(pair(
                 alpha1,
-                many0(alt((alphanumeric1, tag("_"), tag("-"), tag("%")))),
+                many0(alt((alphanumeric1, tag("_"), tag("%")))),
```
mmarx, line 319: As far as I can tell, this breaks compatibility with SPARQL, since prefix names may contain `-`: https://www.w3.org/TR/sparql11-query/#rPN_CHARS

aannleax: Hm, but its hard to fix without running into the problem above. Consider the following program:

```
@prefix my-prefix <https://prefix.com> .

result(my-prefix:test) :- body(?x).
```
Now `my-prefix:test` can either be the constant with the prefix `my-prefix` or the arithmetic operation `my <minus> prefix:test`. 

We talked about fixing this sort of ambiguity in the context of having `:` as key-value separator in maps by having the parser state include all known prefixes,. But this would require a more thorough extension of the parser that is not in the scope of this quick bug-fix.

aannleax: I will create an issue about this so to keep track of this

PR #547, opened 2024-10-30
Bugfix/544 error in empty editor

### LINKED ISSUE #544: Web Application: Error when typing into empty editor
In the web application, when starting from an empty program, typing anything leads to the following error message popping up: "Request textDocument/documentSymbol failed."

This is currently only visible in https://tools.iccl.inf.tu-dresden.de/nemo/next/ as I just disabled this error message in https://tools.iccl.inf.tu-dresden.de/nemo

monsterkrampe:
Fixes #544 

This version is already deployed here:
https://tools.iccl.inf.tu-dresden.de/nemo/next
Feel free to double check that the error is gone :)


aannleax (approved):
Looks good :)

PR #548, opened 2024-11-01
Simplify `TableSource`

matzemathics:
removes one level of structures, which is no longer needed


monsterkrampe:
Looks good on first sight :)  
Did you discuss this with anyone offline? If not, then I think it would be best practice to open an issue before to discuss possible intentions of the current solution (which only applies if this was not written by yourself) and also just to keep track that someone is working on this.

monsterkrampe:
I think there would not be much discussion needed in this particular case and creating a bunch of issues might also be overkill indeed. I just want to make sure that the other devs are aware that changes are currently happening in the code base (which other people might be working on simultaneously). This can happen through issues, creating a draft PR early on, or just putting a message in our internal chat I suppose :)

PR #549, opened 2024-11-02
Split `ExecutionEngine::execute` into smaller functions

matzemathics:
note: based on #548

Rule execution is now performed in `ExecutionEngine::step`, which returns a `Vec` of the changed predicates. De fragmentation of tables is performed separately in `ExecutionEngine::defrag`.

Also `ExectionEngine::initialize` takes ownership of its `Program` parameter instead of taking a reference and then cloning it, which is just bad practice.

Finally the odd `ChaseProgram` parameter, which was handed down to the tracing functions is removed, by being a bit more conscious about where and what to clone. Note that this will likely change later, as I'm cleaning up more of the tracing code.

### DIFF
--- a/nemo-cli/src/main.rs
+++ b/nemo-cli/src/main.rs
@@ -282,7 +282,8 @@ fn run(mut cli: CliApp) -> Result<(), CliError> {
         cli.import_directory.clone(),
     ));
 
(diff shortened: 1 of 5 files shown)


aannleax (approved):
Like the change :)

PR #553, opened 2024-12-02
Dictionaries for tuple data values

mkroetzsch:
This branch introduces support for labelled and unlabelled tuple data values to be used in the dictionary.


mkroetzsch:
## Some explanation for potential reviewers

This branch introduces a new form of dispatch to accomplish code separation in child dictionaries in a situation where a child dictionary would need to know (and call) its parent, which rust does not allow to happen due to borrowing rules. The solution is to replace methods of the child dictionary by global functions (provided by the child dictionary) that receive only the parent dictionary and an id that allows the original child dictionary to be retrieved when needed. Some more magic with Any is used to cast these child dictionaries into the types that are required. The default setup for this can be seen in the "with parent" default implementations in `datavalue_dictionary.rs`, which merely call the child dictionary's methods there.

The net effect of this stunt is that we can keep tuple serialisation/deserialisation code in the TupleDvDictionary although this code requires access to the parent dictionary to encode/decode values inside the tuple.

matzemathics (changes requested):

nemo-physical/src/datavalues/any_datavalue.rs
```diff
@@ -839,6 +840,15 @@ impl From<MapDataValue> for AnyDataValue {
     }
 }
 
+impl From<AnyDataValue> for IriDataValue {
```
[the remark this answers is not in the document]
mkroetzsch: Good catch, will fix.

nemo-physical/src/dictionary/datavalue_dictionary.rs
```diff
@@ -65,6 +67,22 @@ pub trait DvDict: Debug + ByteSized {
...
+    /// borrow checker even if self is never accessed directly within the method. Returning a function pointer
+    /// replaces the native dynamic dispatch without requiring `&self` to be present.
+    ///
+    /// The default implementation of this method returns a pointer to a function that
+    /// simply calls [DvDict::add_datavalue] after retrieving the identified dictionary from the parent.
+    /// There is no major performance penalty in this double indirection since [MetaDvDictionary] can cache the
+    /// function pointer, and would have to retrieve the sub-dictionary anyway to call it.
+    fn add_datavalue_with_parent_fn(
+        &self,
+    ) -> fn(&mut MetaDvDictionary, dict_id: usize, dv: AnyDataValue) -> AddResult {
```
matzemathics, line 82: this is a stylistic choice, but you could also implement those function pointers as closures (since they can be converted into fn pointers automatically, as long as they don't capture anything).

nemo-physical/src/dictionary/datavalue_dictionary.rs
```diff
@@ -65,6 +67,22 @@ pub trait DvDict: Debug + ByteSized {
...
+        add_datavalue_with_parent
```
matzemathics, line 83:
```suggestion
       |parent_dict, dict_id, dv| {
            parent_dict
                .sub_dictionary_mut_unchecked(dict_id)
                .add_datavalue(dv)
        }
```

nemo-physical/src/dictionary/datavalue_dictionary.rs
mkroetzsch: What would be the benefit of doing this?

mkroetzsch: Ah, maybe I get it. You want to "inline" the functions in this way. The signature of the method would stay but the functions move "up" in the source file? Maybe one could try that, but the more complex cases will use auxiliary (private) functions that would have to be outside of the trait implementation, in this case leading to larger distance between related code n the source file.

nemo-physical/src/dictionary/datavalue_dictionary.rs
```diff
@@ -115,4 +165,52 @@ pub trait DvDict: Debug + ByteSized {
 
     /// Returns true if the dictionary contains any marked elements (see [DvDict::mark_dv]).
     fn has_marked(&self) -> bool;
+
+    /// Returns a mutable object that implements this trait as a [std::any::Any]. Dynamic
+    /// trait objects cannot be used as [std::any::Any], but concrete implementations
+    /// can provide a (dynamic) method that accomplishes this. This is a pattern that
+    /// can then be used for casting the dynamic object into a concrete type.
+    fn as_any_mut(&mut self) -> &mut dyn std::any::Any;
```
matzemathics, line 185: These methods should probably not be part of this trait, since they can trivially be implemented for (almost) any type.

Instead the following is more scalable (no need to implement as_any for every dictionary type):

```
trait AsAny {
  fn as_any_mut(&mut self) -> &mut dyn std::any::Any;
  fn as_any(&self) -> &dyn std::any::Any;
}

impl<T: 'static> AsAny for T {
  fn as_any_mut(&mut self) -> &mut dyn std::any::Any {
    self
  }

  fn as_any(&self) -> &dyn std::any::Any {
    self
  }
}
```

With that the `as_any` method is available for all types where applicable.

matzemathics (changes requested):

nemo-physical/src/dictionary/datavalue_dictionary.rs
mkroetzsch: Ok, so I did this now for the short default implementations in this file. I am not convinced that the same approach would make the longer code in tuple_dv_dict.rs more readable though. One of the callbacks there also has own documentation that describes when it will panic; would probably not fit in front of a closure.

nemo-physical/src/dictionary/datavalue_dictionary.rs
mkroetzsch: Nice solution. Will try.

nemo-physical/src/dictionary/tuple_dv_dict.rs
```diff
@@ -0,0 +1,605 @@
...
+
+impl DvDict for TupleDvDict {
+    /// This method always rejects.
+    fn add_datavalue(&mut self, _dv: AnyDataValue) -> AddResult {
+        AddResult::Rejected
+    }
+
+    fn add_datavalue_with_parent_fn(
+        &self,
+    ) -> fn(&mut MetaDvDictionary, dict_id: usize, dv: AnyDataValue) -> AddResult {
```
matzemathics, line 86: Same closure idea also applies here

mkroetzsch: I did this now for the short default implementations in datavalue_dictionary.rs. I am not convinced that the same approach would make the longer code in tuple_dv_dict.rs more readable though. At least one of the callbacks there also has own documentation that describes when it will panic; such docs would probably not fit in front of a closure, and having this all indented by several more levels seems not so readable to me.

matzemathics (approved):
I think this looks good for now. Not sure how to efficiently solve the `tuple_bytes` vs `tuple_bytes_mut` issue, or if there even is a good solution currently.

PR #554, opened 2024-12-04
Implement `Permutation` using slices

matzemathics:
This is a re-implementation of `Permutation` getting rid of the inner `HashMap`, the main benefit being that `Permutation` can now `derive(Ord)`. This is useful, because it is used as a key in the `ColumnOrderMap`. Also it is probably more memory efficient.


aannleax (approved):
Looks good

PR #561, opened 2024-12-05
Use correct file extensions for TSV and DSV

### LINKED ISSUE #559: TSV and DSV formats use the wrong file extension
Currently `@export Test :- tsv{}.` results in the creation of the new table `Test.csv`.
It would be more clear if the extension was `.tsv`

mmarx:
Fixes #559.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -2,7 +2,7 @@ rec {
   description = "nemo, a datalog-based rule engine for fast and scalable analytic data processing in memory";
 
   inputs = {
-    nixpkgs.url = "github:NixOS/nixpkgs/nixos-23.11";
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-24.11";
     nixpkgs-unstable.url = "github:NixOS/nixpkgs/nixos-unstable";
 
     rust-overlay.url = "github:oxalica/rust-overlay";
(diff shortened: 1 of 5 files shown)


flake.nix
```diff
@@ -25,7 +25,7 @@ rec {
       channels.nixpkgs.overlaysBuilder = channels: [
         rust-overlay.overlays.default
         (final: prev: {
-          inherit (channels.nixpkgs-unstable) wasm-bindgen-cli;
+          #          inherit (channels.nixpkgs-unstable) wasm-bindgen-cli;
         })
```
monsterkrampe, line 29: I there a reason for not removing this altogether?

mmarx: So that we can more easily put it back in, once the version in stable nixpkgs gets inevitably out of date. But if you prefer, yeah, I can also remove this (and the whole `nixpkgs-unstable` input with it).

mmarx: I've moved this into its own commit ([ee0b9d2](https://github.com/knowsys/nemo/pull/561/commits/ee0b9d286296165769f2981230a29fe4a9bf2711)), which can be reverted should the need arise.

monsterkrampe: Thanks!
I would also be fine with leaving it but putting explanation why this is commented out :)

monsterkrampe (approved):

PR #565, opened 2024-12-09
Feature/display name attributes

### LINKED ISSUE #562: Feature: Rule attributes
The syntax should support attributes, inspired by [Rust attributes](https://doc.rust-lang.org/reference/attributes.html), that can be attached to rules.

E.g. one could set a rule name like so:

```
#[name("my rule")]
r(?x) :- a(?x).
```

aannleax:
Implements rule attributes. Right now only support name and display which determines how rules are displayed in the ascii representation of proof trees.

Closes #562


nemo/src/rule_model/error/translation_error.rs
```diff
@@ -95,6 +95,30 @@ pub enum TranslationErrorKind {
...
+    /// Unkown attribute
+    #[error("unknown attribute: `{0}`")]
+    #[assoc(code = 119)]
+    AttributeUnknown(String),
+    /// Unexpected attribute
+    #[error("unexpected attribute: `{0}`")]
+    #[assoc(code = 120)]
+    AttributeUnexpected(String),
+    /// Attributed defined multiple times
+    #[error("")]
```
mmarx, line 107: Should this be empty?

mmarx (approved):

Issue #569, opened 2024-12-14
Preliminary support for reasoning over SPARQL query results


mmarx:
We can already load EDB predicates from external sources. In theory, this could be used to query SPARQL endpoints such as the Wikidata Query Service. In practice, this results in unreadable, hardcoded IRIs and depends on the endpoint to send the results in the correct format.

Building upon #566 and #564, we can improve the ergonomics of such queries by adding a builtin and slightly generalising existing syntax, allowing, e.g.:
```prolog
@import awards :- tsv{ resource=f"https://query.wikidata.org/sparql?query={URIENCODE(?query)}",
                       format=(iri, string)
                     },
               ?query = "SELECT ?award ?awardLabel WHERE { wd:Q42 wdt:P166 ?award . ?award rdfs:label ?awardLabel FILTER(LANG(?awardLabel) = \"en\")"
```

- [x] #571
- [x] #572
- [x] #573
- [x] #574
- [x] #575

PR #577, opened 2025-01-04
Preliminary support for reasoning over SPARQL query results

mmarx:
Addresses #569. Based on #564 and #566, so should be merged after those.

With this, we can run the following program:
```prolog
@import awards :- tsv{ resource=f"https://query.wikidata.org/sparql?query={URIENCODE(?query)}",
                       format=(any, string)
                     },
               ?query = """
    SELECT ?award ?awardLabel WHERE {
      wd:Q42 wdt:P166 ?award .
      ?award rdfs:label ?awardLabel FILTER(LANG(?awardLabel) = "en")
    }
    """.

@export awards :- csv{resource=""} .
```

This currently yields the following output:
```
[2025-01-04T00:15:11Z INFO  nemo::io::formats::dsv::writer] Starting data export
?award,"""?awardLabel"""
http://www.wikidata.org/entity/Q906455,"""Ditmar Award@en"""
http://www.wikidata.org/entity/Q2687578,"""Inkpot Award@en"""
[2025-01-04T00:15:11Z INFO  nemo::io::formats::dsv::writer] Finished export: processed 3 tuples (dropped 0)
```


mkroetzsch:
Regarding `URIENCODE`, does this behave like JS's `encodeuri` or `encodeuricomponent`? (See https://stackoverflow.com/questions/4540753/should-i-use-encodeuri-or-encodeuricomponent-for-encoding-urls) Seems the latter would be needed here.

mmarx:
> Regarding `URIENCODE`, does this behave like JS's `encodeuri` or `encodeuricomponent`?

It's like `encodeURIComponent`: “Percent-encodes every byte except alphanumerics and `-`, `_`, `.`, `~`.”

monsterkrampe (approved):
LGTM :)

mkroetzsch:
Some more questions about this new feature:
1. Does the use of additional variable assignments after import directives also work if there are several such assignments, as in `..., ?a="hello", ?b=STRCONCAT(?a, " world")`?
2. Does this work with all directives now or just with imports?

mkroetzsch:
And another design question: Would it make sense to have global constant definitions (for the whole file) rather than local variable assignments (for each directive)? I can imagine use cases where one would like to use a specific constant in a program, but that constant (e.g., a Wikidata ID) could be changed to any value. The above SPARQL example would also look nicer if we could define the query string first, and then use it in subsequent imports.

nemo/src/parser/ast/directive/import.rs
```diff
@@ -19,10 +22,12 @@ pub struct Import<'a> {
     /// [Span] associated with this node
     span: Span<'a>,
 
-    /// Predicate that is being Imported
+    /// Predicate that is being imported
     predicate: StructureTag<'a>,
     /// Map of instructions
     instructions: Map<'a>,
+    /// Additional variable bindings
+    guards: Option<Sequence<'a, Guard<'a>>>,
```
aannleax, line 30: Would be nice to have a unit test for the new syntax (also in export.rs)

aannleax (changes requested):
Look pretty good :)

mmarx:
> Does the use of additional variable assignments after import directives also work if there are several such assignments, as in `..., ?a="hello", ?b=STRCONCAT(?a, " world")`?

the assignments all need to be of the form `?variable = groundish-term`, where `groundish-term` is either a ground term or something that reduces to one (e.g., `?x = 23 * 42` works). There can be multiple assignments, though.

> Would it make sense to have global constant definitions (for the whole file) rather than local variable assignments (for each directive)? I can imagine use cases where one would like to use a specific constant in a program, but that constant (e.g., a Wikidata ID) could be changed to any value. The above SPARQL example would also look nicer if we could define the query string first, and then use it in subsequent imports.

For now, one could simply use a unary predicate for this. That doesn't protect against providing multiple facts for it, though, which would lead to undesired behaviour, so I can imagine that we want to add constants at some point. In either case, though, I would see this in addition to the constant bindings on import/export, not as an alternative.

aannleax (approved):

matzemathics (approved):

PR #579, opened 2025-01-08
generalize `ProgramAST` lifetimes

matzemathics:
This slightly generalizes the signature of `ProgramAST::children`


mmarx (approved):
LGTM

PR #583, opened 2025-01-08
Refactor rule model translation

matzemathics:


nemo/src/io/formats/dsv.rs
```diff
@@ -104,4 +101,8 @@ impl ImportExportHandler for DsvHandler {
     fn import_export_resource(&self) -> &ImportExportResource {
         &self.resource
     }
+
+    fn file_format(&self) -> FileFormat {
+        FileFormat::DSV
+    }
```
mmarx, line 107: This breaks the media type selection for HTTP requests, which will now send the wrong `Accept:` headers for CSV and TSV.

mmarx (changes requested):

nemo/src/io/formats/dsv.rs
matzemathics: oops, didn't mean to change that

mmarx: Yeah, if I hadn't just introduced it, I'd likely have missed it too. I should have written a test for that :D

nemo/src/rule_model/components/term/operation.rs
```diff
@@ -330,13 +313,24 @@ impl IterablePrimitives for Operation {
...
+    impl Operation {
+        fn parse(input: &str) -> Result<Operation, ComponentParseError> {
+            let Term::Operation(op) = Term::parse(input)? else {
+                return Err(ComponentParseError::ParseError);
+            };
+
+            Ok(op)
+        }
+    }
+
```
mmarx, line 333: Should this really go into `mod test`?

nemo/src/rule_model/translation.rs
```diff
@@ -4,42 +4,51 @@ pub(crate) mod attribute;
...
     /// Label of the input file
     input_label: String,
 
     /// Mapping of [Origin] to [ProgramAST] nodes
-    origin_map: HashMap<Origin, &'a dyn ProgramAST<'a>>,
+    origin_map: HashMap<Origin, &'b dyn ProgramAST<'a>>,
 
-    /// Prefix mapping
-    prefix_mapping: HashMap<String, (String, &'a ast::directive::prefix::Prefix<'a>)>,
+    /// Prefix mapping — 1025
```
mmarx, line 48: what is the meaning of `— 1025` here?

mmarx (approved):
LGTM, but I have a couple of questions

nemo/src/rule_model/translation.rs
matzemathics: I have no idea -- maybe its vs-code starting to gaslight me :sweat_smile: 
```suggestion
    /// Prefix mapping
```

nemo/src/rule_model/components/term/operation.rs
matzemathics: This was the easiest way of getting out of unused code warnings, since `parse` is only used for unit tests.

Issue #573, opened 2024-12-14
HTTP(S) requests for data source should send `Accept` headers


mmarx:
Depending on the data source in use, an appropriate `Accept` header (e.g., `Accept: text/tab-separated-values` for `tsv`, or `Accept: text/csv` for `csv`) should be sent.

PR #584, opened 2025-01-08
Add test case for Media types

mmarx:
As #583 accidentally almost broke the new media type support (#573), add some test cases making sure this doesn't happen again.

Based on #583, shouldn't be merged before.


matzemathics (approved):
Looks great :tada:

PR #586, opened 2025-01-09
Add option to ignore DSV headers

mmarx:
Add an `ignore_headers` attribute to `@import` directives that drops the first record in DSV sources.


matzemathics (approved):
Looks fine. Even a negative diff :+1:

PR #588, opened 2025-01-10
Bug/582 float nan inf

### LINKED ISSUE #582: Bug: Division with extreme numbers crashes
```
Big(1.7976931348623157E308).
Small(1.0E-32).
A(?a / ?b) :-
Big(?a),
Small(?b).
```
produces following error:
```
thread 'main' panicked at nemo-physical/src/datavalues/float_datavalues.rs:115:13:
floating point number must represent a finite value (neither infinity nor NaN are allowed).
```

aannleax:
This PR fixes a crash that happens when the computation result returns infinity. This was caused by a mismatch of the definition for the floating point datavalue and our floating point wrappers for f32 and f64. They should now be aligned.

Fixes #582.


nemo-physical/src/datatypes/double.rs
```diff
@@ -20,30 +30,30 @@ impl Double {
...
     }
 
-    /// Wraps the given [f64]-`value`, that is a number, as a value over [Double].
+    /// Wraps the given [f64]-`value` as a value over [Double].
     ///
     /// # Panics
-    /// The given `value` is [f64::NAN].
+    /// Panics if `value` is [f64::NAN] or not finite.
     pub fn from_number(value: f64) -> Self {
         if value.is_nan() {
```
mmarx, line 47: This should be `!value.is_nan()` as well, I presume from the comment?

nemo-physical/src/datatypes/double.rs
```diff
@@ -20,30 +30,30 @@ impl Double {
...
+            panic!("floating point values must be finite")
         }
 
         Self(value)
     }
 
     /// Computes the absolute value.
     pub(crate) fn abs(self) -> Self {
-        Double::new(self.0.abs()).expect("Taking the absolute value cannot result in NaN")
+        Double::new(self.0.abs()).expect("operation returned in invalid float")
```
mmarx, line 56: at the very least, this should be grammatically correct, i.e., `operation returned an invalid float`.

Furthermore, we've generally been using [expect as precondition](https://doc.rust-lang.org/std/error/index.html#common-message-styles), so it should actually be something like `operation returns finite float`

nemo-physical/src/datatypes/double.rs
```diff
@@ -1,13 +1,23 @@
...
+
+use crate::{
+    error::{Error, ReadingError},
+    function::definitions::numeric::traits::{CheckedPow, CheckedSquareRoot},
+};
+
+use super::{run_length_encodable::FloatingStep, FloorToUsize, RunLengthEncodable};
 
 #[cfg(test)]
 use quickcheck::{Arbitrary, Gen};
```
mmarx, line 23: It also doesn't allow infinity values now, the comment should reflect that.

nemo-physical/src/datatypes/float.rs
```diff
@@ -1,49 +1,58 @@
...
+use crate::{
+    error::{Error, ReadingError},
+    function::definitions::numeric::traits::{CheckedPow, CheckedSquareRoot},
+};
 
 #[cfg(test)]
 use quickcheck::{Arbitrary, Gen};
 
-/// Wrapper for [f32`] that does not allow [`f32::NAN] values.
+/// Wrapper for [f32] that does not allow [f32::NAN] values.
```
mmarx, line 24: The comment should also indicate that it excludes infinity values.

nemo-physical/src/datatypes/float.rs
```diff
@@ -1,49 +1,58 @@
...
+            panic!("floating point values must be finite")
         }
 
         Float(value)
     }
 
     /// Computes the absolute value.
     pub(crate) fn abs(self) -> Self {
-        Float::new(self.0.abs()).expect("Taking the absolute value cannot result in NaN")
+        Float::new(self.0.abs()).expect("operation returned in invalid float")
```
mmarx, line 55: see above.

mmarx (changes requested):
This is mostly about the wording of comments and error messages, but you also missed checking for infinity at one point.

nemo-physical/src/datatypes/double.rs
```diff
@@ -20,30 +30,30 @@ impl Double {
     /// Wraps the given [f64]-`value` as a value over [Double].
```
mmarx, line 30: It also doesn't allow infinity values now, the comment should reflect that.

mmarx (approved):

PR #589, opened 2025-01-13
Feature/570 builtin iri

### LINKED ISSUE #570: `IRI` builtin that turns a string into an IRI
A builtin that turns `IRI("https://example.org")` into `<https://example.org>`.

It turns out this is not actually required for #569, since `resource` takes a `string`, not an `iri`, but it could still be useful to have.

aannleax:
Adds the builtin `IRI` that takes as input a plain string and casts its type to an IRI.

This PR also changes the name of the builtin `INVERSE` for multiplying a number with negative one to `INVERTSIGN` which should clash less with user-chosen names.

Closes #570.

### DIFF
--- a/nemo-physical/src/function/definitions/casting.rs
+++ b/nemo-physical/src/function/definitions/casting.rs
@@ -200,7 +200,7 @@ impl UnaryFunction for CastingIntoDouble {
 
 /// Casting of a string value into an IRI
 ///
-/// Returns am IRI with the same content as the given string.
+/// Returns an IRI with the same content as the given string.
 ///
 /// Returns `None` when called on values other than plain strings.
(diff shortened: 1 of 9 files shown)


nemo-physical/src/function/definitions/casting.rs
```diff
@@ -197,3 +197,27 @@ impl UnaryFunction for CastingIntoDouble {
         FunctionTypePropagation::KnownOutput(StorageTypeName::Double.bitset())
     }
 }
+
+/// Casting of a string value into an IRI
+///
+/// Returns am IRI with the same content as the given string.
```
mmarx, line 203: `Returns an`

mmarx (changes requested):
Found a typo, but looks good otherwise!

PR #591, opened 2025-01-15
apply the default compression format to all uncompressed exports

### LINKED ISSUE #590: Command line `--gzip` option is currently being ignored
This issue seems to be caused by a regression in the `ExportManager` code. Will be fixed together with #585

matzemathics:
closes #590


mmarx (approved):

mmarx:
Probably out of scope for this PR (or ever for 0.7, for that matter), but I guess we should have some integration tests for the interaction of compression formats specified for resources and the `--gzip` argument.

PR #601, opened 2025-01-22
add builtin functions BITSHL, BITSHR, BITSHRU

Laschoking:
All 3 functions take `(value,base)` as arguments. If `base` is negative, all functions will return `None` since negative shifts are not supported.

- `BITSHR` is the arithmetic right shift, meaning that the negation sign is kept (insert `1` as Most Significant Bit if number is negative, insert `0` if number is positive). 
- `BITSHRU` is the logical right shift, meaning that the negation sign is not kept (always insert `0` as Most Significant Bit).
- `BITSHL` shifts left by inserting `0`  as Least Significant Bit

### DIFF
--- a/nemo-physical/src/function/definitions/numeric.rs
+++ b/nemo-physical/src/function/definitions/numeric.rs
@@ -512,10 +512,10 @@ impl BinaryFunction for BitShiftLeft {
         parameter_second: AnyDataValue,
     ) -> Option<AnyDataValue> {
         if let Some(pair) = NumericPair::from_any_pair_cast(&parameter_first, &parameter_second) {
-            return match pair {
+            match pair {
                 NumericPair::Integer(value, base) => numeric_bitwise_shl(value, base),
                 _ => None,
-            };
+            }
         } else {
             None
         }
@@ -537,10 +537,10 @@ impl BinaryFunction for BitShiftRight {
         parameter_second: AnyDataValue,
     ) -> Option<AnyDataValue> {
         if let Some(pair) = NumericPair::from_any_pair_cast(&parameter_first, &parameter_second) {
-            return match pair {
+            match pair {
                 NumericPair::Integer(value, base) => numeric_bitwise_shr(value, base),
                 _ => None,
-            };
+            }
         } else {
             None
         }
@@ -562,10 +562,10 @@ impl BinaryFunction for BitShiftRightUnsigned {
         parameter_second: AnyDataValue,
     ) -> Option<AnyDataValue> {
         if let Some(pair) = NumericPair::from_any_pair_cast(&parameter_first, &parameter_second) {
-            return match pair {
+            match pair {
                 NumericPair::Integer(value, base) => numeric_bitwise_shru(value, base),
                 _ => None,
-            };
+            }
         } else {
             None
         }
--- a/nemo-physical/src/function/definitions/numeric.rs
+++ b/nemo-physical/src/function/definitions/numeric.rs
@@ -511,15 +511,12 @@ impl BinaryFunction for BitShiftLeft {
         parameter_first: AnyDataValue,
         parameter_second: AnyDataValue,
     ) -> Option<AnyDataValue> {
-        if let Some(pair) = NumericPair::from_any_pair_cast(&parameter_first, &parameter_second) {
-            match pair {
-                NumericPair::Integer(value, base) => numeric_bitwise_shl(value, base),
-                _ => None,
-            }
-        } else {
-            None
+        match NumericPair::from_any_pair_cast(&parameter_first, &parameter_second)? {
+            NumericPair::Integer(value, base) => numeric_bitwise_shl(value, base),
+            _ => None,
         }
     }
+
     fn type_propagation(&self) -> FunctionTypePropagation {
         FunctionTypePropagation::KnownOutput(StorageTypeName::Int64.bitset())
     }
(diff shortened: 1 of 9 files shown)


nemo-physical/src/function/definitions/numeric.rs
```diff
@@ -499,6 +500,81 @@ impl NaryFunction for BitXor {
...
+        parameter_second: AnyDataValue,
+    ) -> Option<AnyDataValue> {
+        if let Some(pair) = NumericPair::from_any_pair_cast(&parameter_first, &parameter_second) {
+            return match pair {
+                NumericPair::Integer(value, base) => numeric_bitwise_shru(value, base),
+                _ => None,
+            };
+        } else {
+            None
+        }
```
mmarx, line 571: a lot of the boilerplate could be avoided by using the `?` operator, like this:
```rust
        match NumericPair::from_any_pair_cast(&parameter_first, &parameter_second)? {
            NumericPair::Integer(value, base) => numeric_bitwise_shl(value, base),
            _ => None,
        }
```

nemo-physical/src/function/definitions/numeric.rs
```diff
@@ -499,6 +500,81 @@ impl NaryFunction for BitXor {
...
+    }
+    fn type_propagation(&self) -> FunctionTypePropagation {
```
mmarx, line 573: there should be an empty line between functions

nemo-physical/src/function/definitions/numeric/integer64.rs
```diff
@@ -210,6 +210,51 @@ pub(super) fn numeric_bitwise_xor(parameters: &[i64]) -> Option<AnyDataValue> {
...
+    if base < 0 {
+        return None;
+    }
+
+    // Apply arithmetic right shift, if value is positive
+    let result: i64 = if value > 0 {
+        value >> base
+    } else {
+        // Convert value in unsigned to do unsigned right shift
+        (value as u64 >> base) as i64
```
mmarx, line 252: use explicit conversions here:
```rust
        let unsigned_value = u64::from_ne_bytes(value.to_ne_bytes());
        let unsigned_result = unsigned_value >> base;

        i64::from_ne_bytes(unsigned_result.to_ne_bytes())
```

nemo-physical/src/function/tree.rs
```diff
@@ -837,6 +838,34 @@ where
...
+    /// Evaluates to an integer resulting from performing the bitwise left shift operation
+    /// on the binary representation of its integer subnodes.
+    pub fn bit_shl(left: Self, right: Self) -> Self {
+        Self::Binary {
+            function: BinaryFunctionEnum::BitShiftLeft(BitShiftLeft),
+            left: Box::new(left),
+            right: Box::new(right),
+        }
+    }
+    /// Needs documentation
```
mmarx, line 852: Can we get the documentation? Also an empty line before the comment, please.

nemo-physical/src/function/tree.rs
```diff
@@ -837,6 +838,34 @@ where
...
+    pub fn bit_shru(left: Self, right: Self) -> Self {
+        Self::Binary {
+            function: BinaryFunctionEnum::BitShiftRightUnsigned(BitShiftRightUnsigned),
+            left: Box::new(left),
+            right: Box::new(right),
+        }
+    }
+
+    pub fn bit_shr(left: Self, right: Self) -> Self {
```
mmarx, line 861: Also needs documentation.

nemo-physical/src/function/definitions/numeric.rs
```diff
@@ -499,6 +500,81 @@ impl NaryFunction for BitXor {
...
+        parameter_second: AnyDataValue,
+    ) -> Option<AnyDataValue> {
+        if let Some(pair) = NumericPair::from_any_pair_cast(&parameter_first, &parameter_second) {
+            return match pair {
+                NumericPair::Integer(value, base) => numeric_bitwise_shr(value, base),
+                _ => None,
+            };
+        } else {
+            None
+        }
```
mmarx, line 546: see above

nemo-physical/src/function/definitions/numeric.rs
mmarx, line 571: see above

mmarx (changes requested):

nemo-physical/src/function/definitions/numeric.rs
```diff
@@ -499,6 +500,81 @@ impl NaryFunction for BitXor {
...
+        parameter_second: AnyDataValue,
+    ) -> Option<AnyDataValue> {
+        if let Some(pair) = NumericPair::from_any_pair_cast(&parameter_first, &parameter_second) {
+            match pair {
+                NumericPair::Integer(value, base) => numeric_bitwise_shl(value, base),
+                _ => None,
+            }
+        } else {
+            None
+        }
```
mmarx, line 521: Even if you don't want to get rid of the explict `let` here (for whatever reason), the `if let Some(…) {…} else { None } ` is just noise that could be
```rust
let pair = NumericPair::from_any_pair_cast(&parameter_first, &parameter_second)?;
match pair { … }
```
which doesn't require looking at different branches.

I'd still prefer
```rust
        match NumericPair::from_any_pair_cast(&parameter_first, &parameter_second)? {
            NumericPair::Integer(value, base) => numeric_bitwise_shl(value, base),
            _ => None,
        }
```

nemo-physical/src/function/definitions/numeric/integer64.rs
```diff
@@ -210,6 +210,52 @@ pub(super) fn numeric_bitwise_xor(parameters: &[i64]) -> Option<AnyDataValue> {
...
+        return None;
+    }
+
+    // Apply arithmetic right shift, if value is positive
+    let result: i64 = if value > 0 {
+        value >> base
+    } else {
+        // Convert value in unsigned to do unsigned right shift
+        let x = value as u64 >> base;
+        x.try_into().unwrap()
```
mmarx, line 253: This causes a panic if the result of the shift doesn't fit into `i64`:
```
dummy(0).

result(bitshru, ?R) :- dummy(_), ?R = BITSHRU(-9223372036854775808, 0).

@export result :- csv{resource=""}.
```

mmarx (changes requested):

mmarx (approved):

PR #602, opened 2025-01-22
Bug/598 evaluate expressions facts

### LINKED ISSUE #598: Crash when using expressions in facts
Currently, the following program crashes

```
test(2+3).

@output test.
```

with the following error message.

```
thread 'main' panicked at nemo/src/chase_model/translation/fact.rs:32:17:
internal error: entered unreachable code: invalid program: fact contains non-primitive values
```

It would be nice if constant expressions would be evaluated using the `reduce` function. For non-ground terms an error message should appear.

aannleax:
The problem was that the validate method of Fact was not triggered before, because Facts used the TranslationComponent implementation for `HeadAtom`. This PR adds such an implementation for `Fact` itself, which then also calls validate. Furthermore, this PR adds some related error messages to the validation of facts. Finally, constant expressions appearing in facts are now evaluated.

Closes #598.


nemo/src/rule_model/program.rs
```diff
@@ -294,6 +294,8 @@ impl ProgramComponent for Program {
     where
         Self: Sized,
     {
+        println!("validate");
+
         for import in self.imports() {
```
monsterkrampe, line 299:
```suggestion
    {
        for import in self.imports() {
```

monsterkrampe (changes requested):
LGTM except for the `println` :D

monsterkrampe (approved):
:+1:

PR #604, opened 2025-01-23
Replace howlong

mmarx:
`howlong` was preventing us from building darwin binaries in CI. This replaces `howlong` with a combination of `cpu-time` and `wasmtimer`, allowing us to (hopefully) build darwin releases as well.

Based on #603, so shouldn't be merged before that.


aannleax (approved):
Very nice :)

PR #606, opened 2025-01-27
Fix timing on WASM

mmarx:
Also includes the language server in the release artefacts.


aannleax (approved):
Thanks for the quick update :)

PR #609, opened 2025-02-02
Refactor file io

### LINKED ISSUE #585: Consolidate file-format specific code
Currently there is file-format specific code spread across the logical and chase models. This makes it hard to add support for new import/export formats, as well as adding features to existing formats.

Eventually all format-specific code such as validation of parameters should go into the io module.

matzemathics:
resolves #585


nemo/src/io/formats/rdf.rs
```diff
@@ -122,46 +133,136 @@ impl ImportExportHandler for RdfHandler {
...
+impl FormatParameter<RdfTag> for RdfParameter {
+    fn required_for(&self, tag: RdfTag) -> bool {
+        if matches!(tag, RdfTag::Rdf) {
+            matches!(
+                self,
+                RdfParameter::BaseParamType(StandardParameter::Resource)
+            )
+        } else {
+            false
+        }
```
mmarx, line 166: Am I reading this correctly and, e.g., `RdfTag::Trig` doesn't require the `resource` parameter?

nemo-wasm/src/lib.rs
```diff
@@ -189,32 +189,20 @@ impl ResourceProvider for BlobResourceProvider {
...
-            // We currently do this in Rust after the Blob has been transferred to the WebAssembly, to reuse Nemo's compression logic.
-            // We could also to this on the JavaScript side, see https://developer.mozilla.org/en-US/docs/Web/API/Compression_Streams_API .
-            if let Some(reader) = compression.try_decompression(cursor) {
-                Ok(Some(reader))
-            } else {
-                Err(ReadingError::Decompression {
-                    resource: resource.to_owned(),
-                    decompression_format: compression.to_string(),
-                })
-            }
```
mmarx, line 217: This will require changes to the JS side, right?

mmarx (commented):
Looks good, but I have two questions.

nemo/src/io/formats/rdf.rs
mmarx: also `if matches!(…) { … } else { false }` is just `matches!(…) && …`

matzemathics: yes to both :sweat_smile:

nemo-wasm/src/lib.rs
matzemathics: No, i don't think so as decompression now happens inside the `Import` handler, see nemo/io/formats.rs:213

matzemathics: https://github.com/knowsys/nemo/blob/2a59068f817393fcbbd3320b97420cc9c80fc6d0/nemo/src/io/formats.rs#L213-L214

mmarx: ah, thanks.

nemo/src/io/formats/rdf.rs
```diff
@@ -156,14 +156,11 @@ format_parameter! {
 
 impl FormatParameter<RdfTag> for RdfParameter {
     fn required_for(&self, tag: RdfTag) -> bool {
-        if matches!(tag, RdfTag::Rdf) {
-            matches!(
+        matches!(tag, RdfTag::Rdf)
+            && matches!(
                 self,
                 RdfParameter::BaseParamType(StandardParameter::Resource)
             )
```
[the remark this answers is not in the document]
matzemathics: Because we don't need the extension to guess the format

mmarx (approved):

PR #610, opened 2025-02-03
Rule Set Classification

monsterkrampe:
This PR will add membership checks for various syntactic (and some semantic) classes of rule sets. The implementation is done by @xR0xEr.  
(This will still take some time and I am mainly creating this Draft PR now to have a place for review and discussion.)


nemo/src/static_checks/rules_properties.rs
```diff
@@ -0,0 +1,199 @@
...
+    fn is_shy(&self) -> bool {
+        let attacked_pos_by_existential_rule_idx_vars: PositionsByRuleIdxVariables =
+            self.attacked_positions_by_existential_rule_idx_variables();
+        println!("{:?}", attacked_pos_by_existential_rule_idx_vars);
+        self.iter()
+            .all(|rule| rule.is_shy(&attacked_pos_by_existential_rule_idx_vars))
+    }
+
+    fn is_mfa(&self) -> bool {
+        todo!("IMPLEMENT");
```
monsterkrampe, line 166: You can use `unimplemented!()`.

nemo/tests/test.rs
```diff
@@ -0,0 +1,446 @@
...
+    };
+
+    // override_exports(&mut program, cli.output.export_setting);
+    // log::info!("Rules parsed");
+    let rule_set: RuleSet = program.rules().cloned().collect();
+    Ok(rule_set)
+}
+
+fn path_canonicalized(path: &str) -> PathBuf {
+    PathBuf::from_str(path).unwrap().canonicalize().unwrap()
```
monsterkrampe, line 437: In general: use `expect(message)` instead of `unwrap` where `message` explains why this operation cannot fail under normal circumstances.

nemo/tests/test.rs
```diff
@@ -0,0 +1,446 @@
...
+    let program_ast = match nemo::parser::Parser::initialize(
+        &program_content,
+        program_filename.clone(),
+    )
+    .parse()
+    {
+        Ok(program) => program,
+        Err((_program, report)) => {
+            report.eprint()?;
+            return Err(Box::new(Error::new(ErrorKind::Other, "error1")));
```
monsterkrampe, line 410: I think it is ok to now have error handling here and to just `expect` `Ok` from the `Result`s since we run these checks on predefined rule sets, which should always succeed parsing.
(Same below.)

resources/testcases_static_checks/.DS_Store
monsterkrampe, line 1: Don't add MacOS specific stuff to git!
(Feel free to add `.DS_Store` to the root `.gitignore`.)

resources/testcases_static_checks/tests/positive/.DS_Store
monsterkrampe, line 1: Remove mac stuff

nemo/src/static_checks/positions.rs
monsterkrampe, line 1: I think there is a lot to be reworked here. 
For example, in general positions are independent of specific acyclicity checks so the code should be independent too, that is generic in places where `MarkingType` and `AttackingType` are currently hardcoded.
As for the `acyclicity_graphs` I do not really like the approach of introducing various traits on type aliases.
I think I need to play around with this a little bit to see what makes sense here...

nemo/src/static_checks/rule_set.rs
monsterkrampe, line 1: I'll also play around here with possible refactorings.

monsterkrampe: I've done the playing around in `acyclicity_graphs.rs`. I think we can make similar adjustments here.

nemo/src/static_checks/positions.rs
monsterkrampe: I've done the playing around in `acyclicity_graphs.rs`. I think we can make similar adjustments here.

mkroetzsch:
General comment: It would be good to think about the API and UI for running these checks. I have created an issue #611 for this purpose.

nemo/Cargo.toml
```diff
@@ -54,6 +54,8 @@ bytecount = "0.6.8"
 colored = "2"
 
 [dev-dependencies]
+assert_cmd = "2.0"
```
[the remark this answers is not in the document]
xR0xEr: It will be removed. I overlooked it somehow.

nemo/src/lib.rs
```diff
@@ -32,6 +32,8 @@ pub mod execution;
 pub mod rule_model;
 pub mod util;
 
+pub mod static_checks;
```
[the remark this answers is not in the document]
xR0xEr: Right now, this module is implemented on data types of the `rule_model` module. Maybe `rule_model_classification` would be a good name? Also since it is based on the `rule_model` module, wouldn't it be better to make it a submodule of `rule_model`?

nemo/src/rule_model/components/rule.rs
```diff
@@ -147,6 +147,24 @@ impl Rule {
         result
     }
 
+    pub fn positive_variables_as_vec(&self) -> Vec<&Variable> {
```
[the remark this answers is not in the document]
xR0xEr: Yes, it would make sense. I used it to get the join variables of a rule. Since same elements in a set is not possible, the other method was not sufficient. Maybe this can be added to the `IterableVariables`-Trait?

nemo/src/static_checks/acyclicity_graph_constructor.rs
```diff
@@ -0,0 +1,25 @@
...
+/// This Trait provides methods of some RuleSet to build the (WeakAcyclicityGraph /
+/// JointAcyclicityGraph).
+pub trait AcyclicityGraphConstructor<'a> {
+    /// Builds the JointAcyclicityGraph.
+    fn joint_acyclicity_graph(&'a self) -> JointAcyclicityGraph<'a>;
+    /// Builds the WeakAcyclicityGraph.
+    fn weak_acyclicity_graph(&'a self) -> WeakAcyclicityGraph<'a>;
+}
+
+impl<'a> AcyclicityGraphConstructor<'a> for RuleSet {
```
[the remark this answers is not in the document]
xR0xEr: I made this because `build_graph` method of `JointAcyclicityGraph` and `WeakAcyclicityGraph` looked pretty much the same. I hoped to refactor that into actually just one method but I couldn't do it myself. The idea was to give an input (in form of an enum) to decide whether the `JointAcyclicityGraph` or `WeakAcyclicityGraph` should be build.

nemo/src/static_checks/acyclicity_graphs.rs
```diff
@@ -0,0 +1,542 @@
...
+
+/// This Trait provides a method to build the (JointAcyclicityGraph / WeakAcyclicityGraph) of a
+/// RuleSet.
+pub trait AcyclicityGraphBuilder<'a> {
+    /// Builds the (JointAcyclicityGraph / WeakAcyclicityGraph) of a RuleSet.
+    fn build_graph(rule_set: &'a RuleSet) -> Self;
+}
+
+impl<'a> AcyclicityGraphBuilder<'a> for JointAcyclicityGraph<'a> {
+    fn build_graph(rule_set: &'a RuleSet) -> JointAcyclicityGraph<'a> {
```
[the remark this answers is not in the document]
xR0xEr: As I said in one of the comments above: The idea was to generalise the `build_graph` method.
Also I don't quite like the idea of just calling this method `new`. Tell me if I am wrong but `new` functions are to get an empty instance or an instance based on input parameters. But here we have to actually build our graph based on our rule set. That's why I would like to stay with this name.

nemo/src/static_checks/acyclicity_graphs.rs
```diff
@@ -0,0 +1,542 @@
...
+    AllPositivePositions, Attacked, AttackedVariables, ExistentialVariables,
+    ExistentialVariablesPositions, RuleIdxRule, RuleIdxVariable, RuleIdxVariables, RulePositions,
+    RuleSet, SpecialPositionsConstructor, Variables,
+};
+use petgraph::algo::is_cyclic_directed;
+use petgraph::graphmap::{DiGraphMap, NodeTrait};
+use std::collections::HashSet;
+
+/// Type to declare a JointAcyclicityGraph of some RuleSet.
+pub type JointAcyclicityGraph<'a> = DiGraphMap<RuleIdxVariable<'a>, ()>;
```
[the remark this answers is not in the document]
xR0xEr: Wrapper structs sound good to me. I actually implemented them before but I thought type alias would also be a good option. I think this can be redone easily.

nemo/src/static_checks/collection_traits.rs
```diff
@@ -0,0 +1,122 @@
...
+            self.insert(item.clone());
+        })
+    }
+
+    fn insert_all_ret(mut self, other: &HashSet<T>) -> HashSet<T> {
+        self.insert_all(other);
+        self
+    }
+
+    fn insert_all_take(&mut self, other: HashSet<T>) {
```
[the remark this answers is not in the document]
xR0xEr: Will be done.

nemo/src/static_checks/rule_properties.rs
```diff
@@ -0,0 +1,205 @@
...
+use crate::rule_model::components::{rule::Rule, IterablePrimitives};
+use crate::static_checks::positions::{Positions, PositionsByRuleIdxVariables};
+use crate::static_checks::rule_set::{
+    AffectedVariables, Attacked, AttackedGlutVariables, AttackedVariables, ExistentialVariables,
+    FrontierVariablePairs, FrontierVariables, GuardedForVariables, JoinVariables, RuleAppearance,
+    RuleAppearancePair, UniversalVariables, Variables,
+};
+
+/// This trait gives some static checks for some rule.
+pub trait RuleProperties {
```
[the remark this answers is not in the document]
xR0xEr: Will be done.

nemo/tests/test.rs
xR0xEr: Will be done.

resources/testcases_static_checks/.DS_Store
xR0xEr: Sorry, I somehow forgot to add it.
Will be done.

nemo/src/static_checks/rule_set.rs
xR0xEr: As you wrote in some comment above some things are unnecessary complicated. I agree on that. Some things can be written more easily and shouldn't need as much code as it does right now.
To be honest, I do not like every refactoring you have done in the last commit. For example, the new implementation of the `new` function of both graphs is written much shorter than before which is good. But I think it cannot be read as easy as before since both functions are quite big now.
Is there a way where we can find a compromise on that one?

nemo/src/lib.rs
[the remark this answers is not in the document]
monsterkrampe: I think I would keep it separate. Maybe even have it in its own crate. These static analysis are useful but not required for anything happening in the Nemo Core so I rather see it as something that we put "on top". But this should be discussed in a bigger group. Some of the checks could potentially be used to influence the reasoning strategies...

nemo/src/static_checks/rule_set.rs
```diff
@@ -0,0 +1,571 @@
...
+#[derive(Debug)]
+pub struct RuleSet(pub Vec<Rule>);
+
+/// Type to relate 2 Variable(s) together.
+#[derive(Debug, Eq, Hash, PartialEq)]
+pub struct RuleAndVariablePair<'a>(pub [RuleAndVariable<'a>; 2]);
+
+// NOTE: KEEP?
+/// This Trait provides methods to get all (HashSet<Position> / Positions) of some type.
+pub trait AllPositivePositions<'a> {
```
monsterkrampe, line 38: Just because a method is implemented on two different structures does not mean that there needs to be a trait for this. 
For this trait and also others in this file I would say that the functionality is too specific to justify generalizing this into a trait.
If you want to keep the trait, it needs to describe a proper capability.

Also `PositivePosition` should be renamed to something like `PositiveLiteralPosition` because it is not clear what it means for a `Position` to be positive.

monsterkrampe (commented):
Looks nice overall. I left some specific comments again :)

nemo-physical/src/datavalues/any_datavalue.rs
```diff
@@ -82,6 +82,16 @@ enum AnyDataValueEnum {
 pub struct AnyDataValue(AnyDataValueEnum);
 
 impl AnyDataValue {
+    // TODO: DELETE FUNCTION ONCE MFA CHECK CAN BE IMPLEMENTED DIFFERENTLY
```
monsterkrampe: Is this still relevant? If so, the comment should give more details about what the requirements are for implementing MFA "differently" and what "differently" means in the first place.

Also, can this move out of this crate and be monkey patched in `nemo-static-checks` (if still required)?

nemo-physical/src/datavalues/string_datavalue.rs
```diff
@@ -16,6 +16,11 @@ impl StringDataValue {
     pub(crate) fn new(string: String) -> Self {
         StringDataValue(string)
     }
+
+    // TODO: DELETE FUNCTION ONCE MFA CHECK CAN BE IMPLEMENTED DIFFERENTLY
+    pub(crate) fn plain_string(&self) -> &String {
+        &self.0
+    }
```
monsterkrampe: Why is `lexical_value` or `to_plain_string_unchecked` not enough? Or in other words: why do you need a reference to the string and not its value?

Cargo.toml
```diff
@@ -4,33 +4,35 @@ default-members = [
     "nemo",
     "nemo-cli",
     "nemo-physical",
-    "nemo-python",
+    #"nemo-python",
```
monsterkrampe, line 7: I think this should be either kept or removed entirely instead of just commented out. (Also below.)

resources/testcases_static_checks/tests_common_data/.DS_Store
monsterkrampe, line 1: MacOS specific files should not be committed. Maybe this was added before the `.gitignore` had been adjusted?

resources/testcases_static_checks/tests/positive/isDmfc/r14.rlsd
monsterkrampe, line 1: This file is called `.rlsd` but links to `.rls` file. This is wrong in a couple of places, sometimes also the other way around. Please double check and fix this.
Also, am I right to assume that the `d` stands for "disjunctive"?

nemo/Cargo.toml
monsterkrampe, line 1: This file should not change. (Also, I'm in favor of trailing newlines in general.)

nemo/src/rule_model/programs/program.rs
```diff
@@ -42,6 +42,15 @@ impl Program {
         }
     }
 
+    /// Returns all rules of the program
+    // TODO: NEEDED FOR TESTS OF STATIC CHECKS, SHOULD BE REMOVED
```
monsterkrampe: So, remove it :)
Or at least move it to the static checks crate.

If it still needed, please elaborate what need to change for it to be removable.

nemo/src/rule_model/pipeline/transformations.rs
```diff
@@ -1,13 +1,15 @@
 //! This module defines [ProgramTransformation]s.
 
 pub mod active;
+pub mod crit_instance;
```
monsterkrampe, line 4: Both, the `crit_instance` and `msa` modules should not be part of the nemo crate but the static checks crate.

nemo/src/rule_model/pipeline/commit.rs
```diff
@@ -38,7 +38,7 @@ pub struct ProgramCommit {
 
 impl ProgramCommit {
     /// Create a new [ProgramCommit] representing an empty program.
-    pub(crate) fn empty(pipeline: Rc<ProgramPipeline>, report: ValidationReport) -> Self {
+    pub fn empty(pipeline: Rc<ProgramPipeline>, report: ValidationReport) -> Self {
```
monsterkrampe, line 41: Double check if the visibility change is still necessary. If it is, then it is ok to make this change :)
(I think nobody really made an effort to think about a good API for the `nemo` crate thus far. Therefore the visibilities might indeed be too strict right now.)

nemo/src/rule_model/pipeline/transformations/skolem.rs
```diff
@@ -62,17 +65,31 @@ impl ProgramTransformation for TransformationSkolemize {
...
                             self.skolem_count += 1;
+                            count_by_ex_vars.insert(ex_var.clone(), self.skolem_count);
                             let name = format!("_SKOLEM_{}", self.skolem_count);
                             *term =
                                 Term::from(FunctionTerm::new(&name, frontier_variables.clone()));
                         }
                     }
                 }
+
+                commit.add_rule(new_rule);
```
monsterkrampe: This looks like a fix for an existing transformation. If this is indeed the case, please submit a separate PR with the fix including at least a regression test.

nemo-static-checks/Cargo.toml
```diff
@@ -0,0 +1,39 @@
+[package]
+name = "nemo-static-checks"
+description.workspace = true
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+homepage.workspace = true
+license.workspace = true
+readme.workspace = true
```
monsterkrampe, line 9: A separate readme for this crate would be desirable.

nemo/src/execution/execution_engine.rs
```diff
@@ -434,6 +434,25 @@ impl<Strategy: RuleSelectionStrategy> ExecutionEngine<Strategy> {
...
+    //
+    //     let number_of_rules: usize = self.program.rules().len();
+    //
+    //     for index in 0..number_of_rules {
+    //         let updated_predicates = self.step(index, &rule_execution[index])?;
+    //         self.defrag(updated_predicates)?;
+    //     }
+    //
+    //     Ok(())
+    // }
```
monsterkrampe: This can simply be removed.

nemo/src/rule_model/components/term/primitive/ground.rs
```diff
@@ -84,6 +84,11 @@ impl GroundTerm {
     pub fn set_value(&mut self, value: AnyDataValue) {
         self.value = value;
     }
+
+    // TODO: DELETE FUNCTION ONCE MFA CHECK CAN BE IMPLEMENTED DIFFERENTLY
```
monsterkrampe: Again, it is unclear what "differently" means and what would be required for this.

nemo/src/rule_model/components/term/aggregate.rs
```diff
@@ -247,6 +247,8 @@ impl PartialEq for Aggregate {
     }
 }
 
+impl Eq for Aggregate {}
```
monsterkrampe: This could also be part of a separate PR. Maybe there could be one that adds all the missing trait implementations you introduced. (As long as they are sensible.)

nemo/src/rule_model/components/term/function.rs
```diff
@@ -97,6 +97,11 @@ impl FunctionTerm {
         self.terms.iter_mut()
     }
 
+    /// Return an iterator over the arguments of this functino term.
+    pub fn into_terms(self) -> impl Iterator<Item = Term> {
+        self.terms.into_iter()
+    }
+
```
monsterkrampe: Use the `terms` method instead. Possibly with the `cloned()` Iterator method. I hope that this is sufficient for your use case.

nemo/src/rule_model/components/term/function.rs
```diff
@@ -323,6 +328,17 @@ impl IterablePrimitives for FunctionTerm {
...
+impl From<(&Tag, Vec<Term>)> for FunctionTerm {
+    fn from((predicate, terms): (&Tag, Vec<Term>)) -> Self {
+        Self {
+            origin: Origin::Created,
+            id: ProgramComponentId::default(),
+            tag: predicate.clone(),
+            terms,
+        }
+    }
+}
```
monsterkrampe: Better use the `new` function instead of implementing this trait.

nemo/src/rule_model/components/atom.rs
```diff
@@ -94,6 +96,16 @@ impl Atom {
         self.predicate.clone()
     }
 
+    /// Return the predicate as a reference of this atom.
+    pub fn predicate_ref(&self) -> &Tag {
+        &self.predicate
+    }
```
monsterkrampe, line 100: Why can you not use the `predicate` method instead? (This question basically applies to all the new methods that returns refs instead of cloned values.)

nemo/src/rule_model/components/atom.rs
```diff
@@ -94,6 +96,16 @@ impl Atom {
...
+
+    /// Return the predicate of this atom as a reference with its length.
+    pub fn predicate_ref_and_len(&self) -> (&Tag, usize) {
```
monsterkrampe: This looks too specific for the basic atom methods. I also don't think that this has much benefit over just calling both individual methods in the places where this is required. (This similarly holds in other places.)

nemo/src/rule_model/components/atom.rs
```diff
@@ -146,6 +158,22 @@ impl DerefMut for Atom {
...
+        self.variables().filter(|var| var.is_universal()).collect()
+    }
+}
+
+/// This Impl-Block contains a method for an atom to get its variables as a reference.
+impl Atom {
+    /// Returns all the Variables of an Atom.
+    pub fn variables_refs(&self) -> HashSet<&Variable> {
+        self.variables().collect()
+    }
```
monsterkrampe: Why? The exposed method already gives you an iterator that you can `collect` into anything you want in the usage sights. I don't see a point in wrapping this into a separate method. Am I missing something?

nemo/src/rule_model/components/atom.rs
```diff
@@ -146,6 +158,22 @@ impl DerefMut for Atom {
     }
 }
 
+/// This Impl-Block contains a method for an atom to get its universal variables.
+impl Atom {
+    /// This method returns the universal Variables of an Atom.
+    pub fn universal_variables(&self) -> HashSet<&Variable> {
+        self.variables().filter(|var| var.is_universal()).collect()
+    }
```
monsterkrampe: This method looks sensible but it should have the same signature as the `variables` method. That is, it should simply return an Iterator. 
This could be part of the separate PR where missing trait implementations are added.

nemo/src/rule_model/components/atom.rs
```diff
@@ -270,6 +298,17 @@ impl IterablePrimitives for Atom {
...
+impl From<(&Tag, Vec<Term>)> for Atom {
+    fn from((predicate, terms): (&Tag, Vec<Term>)) -> Self {
+        Self {
+            origin: Origin::Created,
+            id: ProgramComponentId::default(),
+            predicate: predicate.clone(),
+            terms,
+        }
+    }
+}
```
monsterkrampe: Remove this. Use `new` instead.

nemo-static-checks/src/static_checks/cyclicity_checks/acyclicity.rs
```diff
@@ -0,0 +1,258 @@
...
+    Assignment, CoreReasoner, Cyclic, CyclicityStrategy, FactsByPred, NoBlockStrategy,
+    StrategySelector, Trigger, VarPerAtomIdxPosIdxPerRule, assignments_for_facts,
+    backtrack_sk_term, body_for_assignment, build_var_index_for_rule, build_var_index_for_rules,
+    predicates_ref, reverse_sk, union,
+};
+use crate::static_checks::rule_set::RuleSet;
+
+use std::collections::{HashMap, HashSet};
+
+pub fn mfa_handle(handle: ProgramHandle) -> ProgramHandle {
```
monsterkrampe: Should this not be `mfa_transformation` of something like that?

nemo-static-checks/src/static_checks/cyclicity_checks/acyclicity.rs
```diff
@@ -0,0 +1,258 @@
...
+            filter_new_facts(&mut special_reasoning_set, self.datalog_pr);
+            datalog_reasoner.run_saturating(special_reasoning_set);
+            facts_after_reasoning = union(facts_after_reasoning, datalog_reasoner.into_facts());
+        }
+
+        Trigger::new(trig.rule(), &renamed_ass).is_obsolete(&facts_after_reasoning)
+    }
+}
+
+fn convert_set_to_map(facts: Vec<&Fact>) -> FactsByPred<'_> {
```
monsterkrampe: Could this be a `From` implementation on `FactsByPred`?

nemo-static-checks/src/static_checks/cyclicity_checks/acyclicity.rs
```diff
@@ -0,0 +1,258 @@
...
+    let strat: &dyn CyclicityStrategy = match strat_sel {
+        StrategySelector::MFA => &MFAStrategy,
+        StrategySelector::RMFA => &RMFAStrategy::new(
+            &existential_rules,
+            &datalog_rules,
+            &datalog_pr,
+            &all_pr,
+            &var_per_atom_idx_pos_idx_per_rule,
+        ),
+        _ => unreachable!(),
```
monsterkrampe: I don't see why this would be unreachable.

nemo-static-checks/src/static_checks/collection_traits.rs
```diff
@@ -0,0 +1,150 @@
+//! Functionality that provides generalised methods for collections.
+use std::collections::HashSet;
+use std::hash::Hash;
+use std::iter::IntoIterator;
+
+/// This Trait provides methods for collections to insert every element of another collection.
+pub trait InsertAll<C, T> {
```
monsterkrampe: `T` does not seem to be used at all here.

nemo-static-checks/src/static_checks/collection_traits.rs
```diff
@@ -0,0 +1,150 @@
...
+    fn insert_all(&mut self, other: &C);
+    /// Inserts all elements of another collection into the current collection and returns it.
+    fn insert_all_ret(self, other: &C) -> Self;
+    /// Inserts all elements of a consumed collection into the current collection.
+    fn insert_all_take(&mut self, other: C);
+    /// Inserts all elements of a consumed collection into the current collection and returns it.
+    fn insert_all_take_ret(self, other: C) -> Self;
+}
+
+// NOTE: HASH NEEDED FOR T?
```
monsterkrampe: Can the comment be removed?

nemo-static-checks/src/static_checks/collection_traits.rs
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    fn insert_all_take_ret(mut self, other: Vec<T>) -> HashSet<T> {
+        self.insert_all_take(other);
+        self
+    }
+}
+
+/// This Trait provides methods for collections to remove every element of another collection.
+#[allow(dead_code)]
```
monsterkrampe: If this is never used, then please remove it.

nemo-static-checks/src/static_checks/collection_traits.rs
```diff
@@ -0,0 +1,150 @@
...
+pub trait Superset {
+    /// Checks if there is a superset relation between two instances for some type.
+    fn is_superset(&self, other: &Self) -> bool;
+}
+
+/// This Trait provides a method to look for a disjoint relation between two instances for some type.
+pub trait Disjoint {
+    /// Checks if there is a disjoint relation between two instances for some type
+    fn is_disjoint(&self, other: &Self) -> bool;
+}
```
monsterkrampe, line 148: Are these traits used?

nemo-static-checks/src/static_checks/positions.rs
```diff
@@ -0,0 +1,482 @@
...
+    }
+}
+
+/// This Impl-Block contains a method to create new positions.
+impl Positions<'_> {
+    /// Creates new positions.
+    pub fn new() -> Self {
+        Self(HashMap::<&Tag, HashSet<Index>>::new())
+    }
+}
```
monsterkrampe: Isn't this the same as the derived implementation of `Positions::default()`?

nemo-static-checks/src/static_checks/positions.rs
```diff
@@ -0,0 +1,482 @@
...
+/// Enum to distinguish between (all existential variables / existential variables that appear in
+/// cycles in the joint acyclicity graph) to be the attacking variables.
+#[derive(Clone, Copy, Debug)]
+pub enum AttackingType {
+    /// The attacking variables will be all existential variables.
+    Cycle,
+    /// The attacking variables will be the existential variables that appear in cycles in the
+    /// joint acyclicity graph.
+    Existential,
+}
```
monsterkrampe: The doc comments do not seem to match the names.

nemo-static-checks/src/static_checks/positions.rs
```diff
@@ -0,0 +1,482 @@
...
+    fn conclude_attacked_positions(&'a self, cur_att_pos: &Positions<'a>) -> Positions<'a> {
+        self.0
+            .iter()
+            .fold(Positions::new(), |new_found_attacked_positions, rule| {
+                let new_att_pos_in_rule: Positions = rule.conclude_attacked_positions(cur_att_pos);
+                new_found_attacked_positions.insert_all_take_ret(new_att_pos_in_rule)
+            })
+    }
+
+    // TODO: INSTEAD OF POSITIONS_IN_HEAD CALL, CALL INIT_ATT_POS OF RULE
```
monsterkrampe: Should this be addressed?

nemo-static-checks/src/static_checks/positions.rs
```diff
@@ -0,0 +1,482 @@
...
+            .map(|var| RuleAndVariable(self, var))
+            .filter(|rule_and_var| rule_and_var.is_attacked_by_positions(cur_att_pos))
+            .fold(Positions::new(), |new_att_pos_in_rule, rule_and_var| {
+                let pos_of_var_in_head: Positions = rule_and_var.positions_in_head();
+                new_att_pos_in_rule.insert_all_take_ret(pos_of_var_in_head)
+            })
+    }
+
+    fn initial_attacked_positions(&'a self, _rule_and_var: &RuleAndVariable<'a>) -> Positions<'a> {
+        todo!("IMPLEMENT");
```
monsterkrampe: Please do :D

nemo-static-checks/src/static_checks/rule_properties.rs
```diff
@@ -0,0 +1,209 @@
...
+    fn is_glut_frontier_guarded(
+        &self,
+        attacked_pos_by_cycle_rule_and_vars: &PositionsByRuleAndVariables,
+    ) -> bool {
+        let attacked_frontier_glut_variables: HashSet<&Variable> =
+            self.attacked_frontier_glut_variables(attacked_pos_by_cycle_rule_and_vars);
+        self.is_guarded_for_variables(attacked_frontier_glut_variables)
+    }
+
+    // TODO: SHORTEN FUNCTION
```
monsterkrampe: I thinks it's ok.

nemo-static-checks/src/static_checks/rule_properties.rs
```diff
@@ -0,0 +1,209 @@
...
+                attacked_pos_by_rule_and_vars.0.values().all(|ex_var_pos| {
+                    !rule_and_var_1.is_attacked_by_positions(ex_var_pos)
+                        || !rule_and_var_2.is_attacked_by_positions(ex_var_pos)
+                })
+            });
+        join_vars_in_multiple_body_atoms_are_not_attacked
+            && frontier_vars_that_appear_in_diff_body_atoms_are_not_att_by_same_var
+    }
+
+    // fn is_mfa(&self) -> bool {
```
monsterkrampe: Remove the commented lines please.

nemo-static-checks/src/static_checks/rule_set.rs
```diff
@@ -0,0 +1,886 @@
...
+use std::collections::{HashMap, HashSet};
+
+/// Type to relate an (existential) Variable to a Rule. Therefore a type to identificate an
+/// (existential) Variable.
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub struct RuleAndVariable<'a>(pub &'a Rule, pub &'a Variable);
+
+impl Ord for RuleAndVariable<'_> {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.1.cmp(other.1)
```
monsterkrampe: The `Ord` implementation is inconsistent with the derived `Eq` implementation!

nemo-static-checks/src/static_checks/rule_set.rs
```diff
@@ -0,0 +1,886 @@
...
+//     pub fn positions_of_existential_variables(&self) -> Positions {
+//         self.existential_rule_and_variables().into_iter().fold(
+//             Positions::new(),
+//             |pos_of_ex_var, rule_and_var| {
+//                 let pos_of_ex_var_in_head: Positions = rule_and_var.positions_in_head();
+//                 pos_of_ex_var.insert_all_take_ret(pos_of_ex_var_in_head)
+//             },
+//         )
+//     }
+// }
```
monsterkrampe: Remove

nemo-static-checks/src/static_checks/rule_set.rs
```diff
@@ -0,0 +1,886 @@
...
+
+pub trait ExistentialVariables {
+    /// Returns all the existential Variables of a rule.
+    // fn existential_variables(&self) -> HashSet<&Variable>;
+    /// Returns all existential variables combined with its rule of a rule.
+    fn existential_rule_and_variables(&self) -> HashSet<RuleAndVariable>;
+    /// Returns the positions of the existential Variables of the rule as a set.
+    fn positions_of_existential_variables_as_set(&self) -> HashSet<Position>;
+    /// Returns the Positions of the existential Variables of a rule.
+    fn positions_of_existential_variables(&self) -> Positions;
```
monsterkrampe: Why both? Aren't there `From` implementations that allow conversion in both ways anyway?

nemo-static-checks/src/static_checks/rule_set.rs
```diff
@@ -0,0 +1,886 @@
...
+            > 1
+    }
+
+    fn appears_without_other_in_positive_body_atom(&self, other: &RuleAndVariable) -> bool {
+        self.0
+            .body_positive_refs()
+            .iter()
+            .filter(|atom| self.1.appears_in_atom(atom) && !other.1.appears_in_atom(atom))
+            .count()
+            > 0
```
monsterkrampe: Could be a call to `any`, no?

nemo-static-checks/src/static_checks/rule_set.rs
```diff
@@ -0,0 +1,886 @@
...
+/// This Impl-Block contains methods for a variable to get its positions in some part of a rule.
+impl<'a> RuleAndVariable<'a> {
+    /// Returns the positions of a variable in the head of some rule as a set.
+    pub fn positions_in_positive_body_as_set(self) -> HashSet<Position<'a>> {
+        let body_pos_of_var: Positions = self.positions_in_positive_body();
+        HashSet::<Position>::from(body_pos_of_var)
+    }
+
+    /// Returns the positions of a variable in the positive body of a rule as a set.
+    pub fn positions_in_head_as_set(self) -> HashSet<Position<'a>> {
```
monsterkrampe: Again, I do not see a use in having extra `as_set` versions. Even if you must have them, they could also have default implementations in the trait.
Please change this in all applicable places.

nemo-static-checks/src/static_checks/rule_set.rs
```diff
@@ -0,0 +1,886 @@
...
+/// This Impl-Block contains a method for a variable pair to check if its variables appear in
+/// different positive body atoms of a rule.
+impl RuleAndVariablePair<'_> {
+    // TODO: REEVALUATE FUNCTION BECAUSE '2 FRONTIER VARIABLE(S) APPEAR IN DIFFERENT ATOM(S)' IS
+    // NOT A CLEAR DEFINITION
+    /// Checks if some Variables of a VariablePair appear at different atoms in the positive body
+    /// of some rule.
+    pub fn appear_in_different_positive_body_atoms(&self) -> bool {
+        let found_var1_alone: bool =
+            self.0[0].appears_without_other_in_positive_body_atom(&self.0[1]);
```
monsterkrampe: Maybe give the `RuleAndVariablePair` accessor functions `fst` and `snd`? This looks unnecessarily complicated right now.
Also, why is the `RuleAndVariablePair` a vector of size two instead of just a tuple struct with two members? (Sorry for this comment being a bit out of place but this just came to my mind.)

nemo-static-checks/src/static_checks/rules_properties.rs
```diff
@@ -0,0 +1,347 @@
...
+            .iter()
+            .all(|rule| rule.is_shy(&attacked_pos_by_existential_rule_and_vars))
+    }
+
+    async fn is_mfa(&self) -> bool {
+        unreachable!();
+    }
+
+    async fn is_msa(&self) -> bool {
+        unreachable!();
```
monsterkrampe: How is this unreachable?! Maybe it's `unimplemented!()`. But why is it? I thought this has been done?

EDIT: Oh I think I see. Some of these only work on a rule set and some only on a program handle? This can not stay as is.

nemo-static-checks/src/cli.rs
```diff
@@ -0,0 +1,46 @@
...
+    /// One or more rule program files
+    #[arg(value_parser, required = true)]
+    pub(crate) rules: Vec<PathBuf>,
+
+    /// One or more static checks
+    #[arg(short = 'c', long = "checks", value_parser = CHECK_NAMES, action = ArgAction::Append, default_values = CHECK_NAMES)]
+    pub(crate) checks: Vec<String>,
+}
+
+const CHECK_NAMES: [&str; 27] = [
```
monsterkrampe: Why is this not an `enum`?

nemo-static-checks/tests/test.rs
```diff
@@ -0,0 +1,486 @@
...
+            exp_res,
+            check,
+        }
+    }
+
+    fn run(&self) -> Result<(), Box<dyn std::error::Error>> {
+        let mut cmd = Command::cargo_bin("nemo-static-checks")?;
+
+        let output = cmd
+            .current_dir("/Users/louis/kbs/nemo/")
```
monsterkrampe: This should not depend on your system!

nemo-static-checks/tests/test.rs
```diff
@@ -0,0 +1,486 @@
...
+            .arg("-c")
+            .arg(self.check)
+            .arg(self.file.as_path())
+            .output()
+            .expect("failed to execute");
+
+        println!("{}", str::from_utf8(&output.stderr).expect("no error"));
+        println!("hello");
```
monsterkrampe: ?

nemo/src/rule_model/components/fact.rs
```diff
@@ -278,6 +278,17 @@ impl IterablePrimitives for Fact {
     }
 }
 
+impl From<(&Tag, Vec<Term>)> for Fact {
```
monsterkrampe: Is this not essentially `Fact::new`?

nemo/src/rule_model/components/literal.rs
```diff
@@ -52,6 +52,24 @@ impl Literal {
             Literal::Operation(_) => None,
         }
     }
+
+    /// If literal is not an operation, return the predicate as a reference.
+    /// Returns `None` otherwise.
+    pub fn predicate_ref(&self) -> Option<&Tag> {
```
monsterkrampe, line 58: As in other places, I do not quite see the need for this method.

nemo/src/rule_model/components/literal.rs
```diff
@@ -52,6 +52,24 @@ impl Literal {
...
+        match self {
+            Literal::Positive(atom) | Literal::Negative(atom) => Some(atom.predicate_ref()),
+            Literal::Operation(_) => None,
+        }
+    }
+
+    /// Return the predicate as a reference and its length (positive Atom | negative Atom) or None (Operation) of the
+    /// Literal
+    pub fn predicate_ref_and_len(&self) -> Option<(&Tag, usize)> {
```
monsterkrampe: Even more so here. The predicate and the arity can be conveniently obtained by two method calls.

nemo/src/rule_model/components/rule.rs
```diff
@@ -106,6 +107,11 @@ impl Rule {
         self.name.clone()
     }
 
+    /// Return an iterator over the terms of the head.
+    pub fn head_terms(&self) -> impl Iterator<Item = &Term> {
+        self.head().iter().flat_map(|atom| atom.terms())
```
monsterkrampe: I do not see a need for having this method. As long as the head can already be retrieved, flat_mapping to the terms can be inlined where it is needed.

nemo/src/rule_model/components/rule.rs
```diff
@@ -164,6 +170,28 @@ impl Rule {
         &mut self.head
     }
 
+    pub fn existential_variables(&self) -> HashSet<&Variable> {
```
monsterkrampe: This should return an iterator. You can always collect this into a HashSet later if needed.
Also below.

nemo/src/rule_model/components/rule.rs
```diff
@@ -367,6 +432,13 @@ impl Rule {
             None
         }
     }
+
+    pub fn contains_func(&self) -> bool {
```
monsterkrampe: This should only exists in the static checks crate.

nemo/src/rule_model/components/term.rs
```diff
@@ -44,7 +44,7 @@ use super::{
 /// Term
 ///
 /// Basic building block for expressions like atoms or facts.
-#[derive(Debug, Clone, PartialEq, Hash, PartialOrd)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd)]
```
monsterkrampe: I do not think that this should implement `Eq`. For example, I think that a term could also have the floating point value `NaN`. But `NaN != NaN` so reflexivity guaranteed by `Eq` would be violated.

nemo/src/rule_model/pipeline/transformations/crit_instance.rs
```diff
@@ -0,0 +1,49 @@
...
+use crate::rule_model::error::ValidationReport;
+use crate::rule_model::programs::handle::ProgramHandle;
+use crate::rule_model::programs::{ProgramRead, ProgramWrite};
+
+use std::collections::HashSet;
+
+#[derive(Debug, Default, Clone, Copy)]
+pub struct TransformationCriticalInstance {}
+
+fn critical_instance(rules: &[&Rule]) -> HashSet<Fact> {
```
monsterkrampe: Are these rules allowed to feature constants? If so, the critical instance needs to involve these constants too, I think. I need to double check the right approach to handle this. Talk to me if constants in rules should be supported.

nemo/src/rule_model/pipeline/transformations/msa.rs
```diff
@@ -0,0 +1,177 @@
...
+use crate::rule_model::programs::handle::ProgramHandle;
+use crate::rule_model::programs::{ProgramRead, ProgramWrite};
+use crate::rule_model::substitution::Substitution;
+
+use std::collections::HashSet;
+
+#[derive(Debug, Default, Clone, Copy)]
+pub struct TransformationMSA {}
+
+fn f_preds_for_ex_vars(ex_vars_of_rule: &HashSet<&Variable>, rule_name: usize) -> Vec<Tag> {
```
monsterkrampe, line 24: I'm confused here. Doesn't MSA simply replace each existential variable by a fresh constant? Why do you need new predicates? What am I missing? (Maybe I'm also misremembering how MSA works.)

nemo-static-checks/src/static_checks/cyclicity_checks.rs
```diff
@@ -0,0 +1,673 @@
...
+                {
+                    Some((*rule, f_term))
+                } else {
+                    None
+                }
+            })
+        })
+        .unwrap();
+    let front_vars = rule.frontier_variables();
+    let mut const_count = 0;
```
monsterkrampe: I think the fresh constants need to be globally unique. This is not ensured here is it? What I mean is, they are only unique within this one rules and not across the rule set, right?

monsterkrampe (commented):
In general, the structure looks good. Don't be discouraged by the number of comments :)

The list of things to address is certainly not exhaustive and I did not verify the checks for correctness. I only had a closer look at some of the details.
Anyway, there should be enough things to address already.
We can (and maybe should) discuss some of this offline of course :)

nemo-physical/src/datavalues/any_datavalue.rs
xR0xEr: Apparently it was not needed anymore. I deleted it.

nemo-physical/src/datavalues/string_datavalue.rs
xR0xEr: Also not needed anymore.

nemo-static-checks/src/static_checks/cyclicity_checks/acyclicity.rs
xR0xEr: I changed the name to `mfa_transformation`, but I think `mfa_handle` should also be fine. Maybe we can discuss offline.

nemo-static-checks/src/static_checks/cyclicity_checks/acyclicity.rs
xR0xEr: Unfortunately, not possible. `FactsByPred` is only a type renaming (`HashMap`) and `From` is a trait from the `std`-crate. Therefore, a trait implementation is not possible.

nemo-static-checks/src/static_checks/cyclicity_checks/acyclicity.rs
xR0xEr: Created separate enums `AcyclicityStrategySelector` and `CyclicityStrategySelector` to distinguish both kinds further. This eliminates this problem.

nemo-static-checks/src/static_checks/collection_traits.rs
xR0xEr: Fixed.

nemo-static-checks/src/static_checks/collection_traits.rs
xR0xEr: Fixed.

nemo-static-checks/src/static_checks/collection_traits.rs
xR0xEr: It was used. dead_code was wrong at this point, but it was unused in some point of time.

nemo-static-checks/src/static_checks/collection_traits.rs
xR0xEr: Yes, they are used. But all three: RemoveAll und Disjoint as well as Superset are only used for the structure Positions. Maybe I should implement it directly on it? But they seem like Trait-like methods.

nemo-static-checks/src/static_checks/cyclicity_checks.rs
xR0xEr: I think this was a mistake from me. I looked over the paper of DRPC / RPC again and this is probably true. I changed it to be globally unique.

nemo-static-checks/src/static_checks/positions.rs
xR0xEr: Yes.

nemo-static-checks/src/static_checks/positions.rs
xR0xEr: I think not. `positions_in_head()` needs a `RuleAndVariable` while `initial_attacked_positions` needs a `Rule` as its first parameter. Therefore, any solution does not seem to be appropriate here and I would rather let this like it is. What do you think?

nemo-static-checks/src/static_checks/rule_set.rs
xR0xEr: I fixed it.

nemo-static-checks/src/static_checks/rule_set.rs
xR0xEr: Yes.

nemo-static-checks/src/static_checks/rule_set.rs
xR0xEr: I changed both to your recommendation. Although, `fst` and `snd` might not seem to be appropriate anymore as the access now is just through `.0` or `.1`, but it is more intuitive, now.

nemo-static-checks/src/cli.rs
xR0xEr: Changed it to an enum.

nemo/src/rule_model/components/term/function.rs
xR0xEr: It is sufficient, but cloning here seems like an unnecessary adjustment.

nemo/src/rule_model/components/term/function.rs
xR0xEr: I can do this, but then I have to make `new_tagged` public. Is this okay?

nemo/src/rule_model/components/atom.rs
xR0xEr: I deleted it.

nemo/src/rule_model/components/atom.rs
xR0xEr: The problem is the following: Datastructures like Positions consist of `&Tag` rather than just `Tag`. Some methods / functions rely on a function give a direct reference to the `Tag`, as it doesn't work to create a reference in a method and returning it (dangling pointer). I could maybe resolve this by just using `Tag` in `Positions`? I thought that I could get a better performance by using a reference.

nemo/src/rule_model/components/atom.rs
xR0xEr: No, you are right. I didn't know what I thought there...

Issue #617, opened 2025-02-05
Integration tests for error messages


matzemathics:
We generate nice user-facing error messages for programs that are invalid (parsing/translation/validation errors).
Ideally we would have a collection of such invalid programs, such that we can ensure those error messages keep being generated as they should.

PR #619, opened 2025-02-07
fix: do not validate facts and rules twice

matzemathics:
this is a regression caused by #609
double validation causes validation errors to appear twice


mmarx:
I trust we'll get a regression test for this with #617 …

mmarx (approved):

matzemathics:
> I trust we'll get a regression test for this with #617 …

Yep, thats how I found it

Issue #618, opened 2025-02-05
Integration testing for remote resources


mmarx:
There should be integration tests verifying, that, e.g., correct HTTP headers are sent when retrieving remote resources. Crucially, these should not require network access to use (which would, e.g., break the nix build, but could also lead to spurious failures), so this requires some kind of embedded web server.

PR #621, opened 2025-02-12
Feature/sparql format

Laschoking:
With this changes a new keyword `sparql` is introduced that expects two keywords:
 `endpoint` is expected to be a valid IRI and `query` a String.
f.e. ` @import test :- sparql{endpoint=<https://query.wikidata.org/>, "SELECT ?item ..."}.` 

Furthermore, resources can be written as IRIs now.
Most importantly the Resource-type is changed, allowing now a `Resource::Iri` for HTTPRequests (either online-files or sparql queries) and `Resource::Path` for local resources.

Known lacks:
- [ ] add documentation
- [ ] integration tests: as pointed out in #618 it is unclear, if integration tests could be evaluated against a real endpoint

Questions for review:
-  how precise should the validation of parameters be? In `sparql::is_value_valid()` the `SparqlParameter::Endpoint` and `SparqlParameter::Query` are validated intensely, which seems  like a lot of effort.
- The parsing of a resource (now can be either IRI or string) was moved to `format::ResouceSpec::parse_string` while before, the two file providers `HttpResourceProvider` and `FileResourceProvider` were handling this. 
-> is this change agreeable?
- in `format_builder::new_with_tag` the created `resource: Option<ResourceSpec>` can be replaced by a `Option<ResourceSpec>` returned from any Builder. This allows the `SparqlBuilder` to return its own ResourceSpec that uses `endpoint` and `query`.
-> is this agreeable or should it be handled elsewhere?
- in `resource_providers::open_resource()` the existing vector containing both `ResourceProviders` is used. -> can this be simplified somehow?


nemo/src/io/resource_providers/http.rs
```diff
@@ -66,26 +75,65 @@ impl HttpResourceProvider {
...
+                // Append the encoded query to URL
+                full_url.push_str("?query=");
+                full_url.push_str(&enc_query);
+                client.get(full_url).send().await.map_err(err_mapping)?
+            }
+        } else {
+            // Dont append any parameters to URL
+            log::debug!("Make GET-request to endpoint {full_url}");
+            client.get(full_url).send().await.map_err(err_mapping)?
+        };
```
mmarx, line 124: Building the target IRI from the `Resource` should happen in `Resource`. The flow for this should be:
 * get the full IRI from the Resource
 * set headers, if applicable
 * if there are any post parameters, do a post request (use `RequestBuilder::post`, so you don't have to worry about correctly encoding them)
 * otherwise, do a get request.

Determining whether the query is done as a get or post request should happen when initially building the resource (it's probably sufficient to check the length of the query for that, we don't really care about browser limits, since these requests will never go through browsers (`nemo-wasm` uses a different API for these requests, and `reqwest` supports larger IRIs, as should the resources), so where exactly the threshould should be is somewhat arbitrary (we can keep 2000, though, it's probably a good default choice).

nemo/src/io/resource_providers/http.rs
```diff
@@ -97,11 +145,14 @@ impl ResourceProvider for HttpResourceProvider {
...
     ) -> Result<Option<Box<dyn Read>>, ReadingError> {
-        if !is_iri(resource) {
-            return Ok(None);
-        }
-
-        if !(resource.starts_with("http:") || resource.starts_with("https:")) {
+        // Add error message
+        let Resource::Iri { iri, parameters } = resource else {
+            unreachable!("The type of the resource is known already")
+        };
```
mmarx, line 151: This needs to go back to what it was before, to support proper selection of Resource providers.

nemo/src/io/resource_providers/http.rs
```diff
@@ -111,7 +162,7 @@ impl ResourceProvider for HttpResourceProvider {
             .build()
             .map_err(|e| ReadingError::from(e).with_resource(resource.clone()))?;
 
-        let response = rt.block_on(Self::get(resource, media_type))?;
+        let response = rt.block_on(Self::get(resource, url, parameters, media_type))?;
```
mmarx, line 165: the `resource` should have all the information, no need to pass `url` and `parameters` here.

nemo/src/rule_model/components/import_export/attributes.rs
```diff
@@ -48,6 +48,16 @@ pub enum ImportExportAttribute {
...
     IgnoreHeaders,
+    /// URL endpoint for SPARQL queries
+    #[assoc(name = attribute::ENDPOINT)]
+    #[assoc(from_name = attribute::ENDPOINT)]
+    #[assoc(value_type = ProgramComponentKind::Iri)]
+    Endpoint,
+    /// The query that is evaluated by the endpoint
+    #[assoc(name = attribute::QUERY)]
+    #[assoc(from_name = attribute::QUERY)]
+    #[assoc(value_type = ProgramComponentKind::PlainString)]
```
mmarx, line 59: We might want to support fstrings here as well (indeed, I'd even prefer another kind of tagged string (`sparql"…"`, which is mostly identical to fstrings, but (i) does syntax validation for SPARQL, and (ii) signals to an editor that it contains something that could be highlighted as SPARQL), but then we should probably have some generalised infrastructur for dealing with all these string variants, so `PlainString` is fine for now.

nemo/src/rule_model/components/import_export/file_formats.rs
```diff
@@ -149,6 +149,20 @@ pub enum FileFormat {
     ]))]
     #[assoc(arity = 4)]
     TriG,
+    /// SPARQL 1.1 
+    /// Is sparql a file format?
```
mmarx, line 153: The answer is no, but treating it as one fits quite well. We should likely refactor this (in a later PR) to `resource_formats`, e.g.?

nemo/src/rule_model/error/validation_error.rs
```diff
@@ -135,32 +135,40 @@ pub enum ValidationErrorKind {
...
     /// Unknown file format
     #[error(r#"unknown file format: `{0}`"#)]
-    #[assoc(code = 231)]
+    #[assoc(code = 233)]
     ImportExportFileFormatUnknown(String),
     /// Unknown arity
     #[error(r#"arity of predicate {predicate} is unknown"#)]
     #[assoc(note = "arity of predicates in import/export statements must be known in advance.")]
-    #[assoc(code = 232)]
+    #[assoc(code = 234)]
```
mmarx, line 171: I'm not sure if we have any formal policy on this, but we should really keep these codes stable (i.e., only add to the end, and, if codes are ever removed, not re-use them again for something different).

nemo/src/io/formats/sparql.rs
```diff
@@ -0,0 +1,186 @@
...
+
+use spargebra::Query;
+use std::sync::Arc;
+
+use nemo_physical::datavalues::{AnyDataValue, DataValue};
+use oxiri::Iri;
+
+//use reader::DsvReader;
+//use value_format::DsvValueFormats;
+//use writer::DsvWriter;
```
mmarx, line 11: Please remove this.

nemo/src/io/formats/sparql.rs
```diff
@@ -0,0 +1,186 @@
+//! Handler for resources of type Sparql (java script object notation).
```
mmarx, line 1: `SPARQL (SPARQL query language for RDF)`

nemo/src/io/formats/sparql.rs
```diff
@@ -0,0 +1,186 @@
...
+        Query(name = attribute::QUERY, supported_types = &[ValueType::String]),
+        IgnoreHeaders(name = attribute::IGNORE_HEADERS, supported_types = &[ValueType::Boolean]),
+
+    }
+}
+
+impl FormatParameter<SparqlTag> for SparqlParameter {
+    fn required_for(&self, tag: SparqlTag) -> bool {
+        if matches!(tag, SparqlTag::Sparql) {
+            matches!(self, SparqlParameter::Endpoint) && matches!(self, SparqlParameter::Query)
```
mmarx, line 50: this needs to be `||`, not `&&`, otherwise it'll never be true. And then you can pull the disjunction into the pattern:
```rust
   matches!(self, Self::Endpoint | Self::Query)
```

mmarx: … and make it 
```rust
  matches!(tag, SparqlTag::Sparql) && matches!(self, Self::Endpoint | Self::Query)
```
to get rid of the `else { false }`

nemo/src/io/formats/sparql.rs
```diff
@@ -0,0 +1,186 @@
...
+                FormatParameter::<SparqlTag>::is_value_valid(base, value)
+            }
+            SparqlParameter::Limit => value
+                .to_u64()
+                .and(Some(()))
+                .ok_or(ValidationErrorKind::ImportExportLimitNegative),
+            SparqlParameter::Base => Ok(()),
+            SparqlParameter::Format => DsvValueFormats::try_from(value).and(Ok(())).map_err(|_| {
+                ValidationErrorKind::ImportExportValueFormat {
+                    file_format: "rdf".to_string(),
```
mmarx, line 70: should be `"sparql"`, I guess?

nemo/src/io/formats/sparql.rs
```diff
@@ -0,0 +1,186 @@
...
+                    })
+                }),
+        }
+    }
+}
+
+#[derive(Clone)]
+pub(crate) struct SparqlBuilder {
+    limit: Option<u64>,
+    delimiter: u8,
```
mmarx, line 99: we certainly can just fix the delimiter to `'\t'`

nemo/src/io/formats/sparql.rs
```diff
@@ -0,0 +1,186 @@
...
+                        }
+                    })
+                }),
+        }
+    }
+}
+
+#[derive(Clone)]
+pub(crate) struct SparqlBuilder {
+    limit: Option<u64>,
```
mmarx, line 98: `limit` would go into the query, wouldn't it?

nemo/src/io/formats/sparql.rs
```diff
@@ -0,0 +1,186 @@
...
+    delimiter: u8,
+    value_formats: Option<DsvValueFormats>,
+    ignore_headers: bool,
```
mmarx, line 101: this also shouldn't be an option, for SPARQL, we *always* want to ignore the heades

nemo/src/io/formats/sparql.rs
```diff
@@ -0,0 +1,186 @@
...
+            parameters
+                .get_required(SparqlParameter::Endpoint.into())
+                .to_iri_unchecked(),
+        );
+
+        let query = parameters
+            .get_required(SparqlParameter::Query.into())
+            .to_plain_string_unchecked();
+
+        let parsed_query = Query::parse(query.as_str(), None).unwrap();
```
mmarx, line 144: This needs proper error handling, otherwise we'll just panic on syntax errors in the query. Generally, `unwrap()` is only ever acceptable in tests, at the very least, there should be an `expect()`.

nemo/src/io/formats/sparql.rs
```diff
@@ -0,0 +1,186 @@
...
+        // Create and return a new ResourceSpec to replace the existing ResourceSpec
+        let resource = ResourceSpec::from_endpoint(endpoint, query);
+        Ok((
+            Self {
+                delimiter: b'\t',
+                value_formats,
+                limit,
+                ignore_headers,
+                parsed_query,
+            },
```
mmarx, line 155: Do we really need to carry all this around? Given the query and the `value_formats`, we have everything we need to just call a `DsvBuilder` once we need it.

nemo/src/io/formats/sparql.rs
```diff
@@ -0,0 +1,186 @@
...
+            | Query::Describe { pattern, .. }
+            | Query::Construct { pattern, .. }
+            | Query::Ask { pattern, .. } => pattern.on_in_scope_variable(|_| var_count += 1),
+        }
+        Some(var_count)
+    }
+
+    fn build_import(&self, arity: usize) -> Arc<dyn ImportHandler + Send + Sync + 'static> {
+        Arc::new(DsvHandler {
+            delimiter: self.delimiter,
```
mmarx, line 173: just hardcode this

nemo/src/io/formats/sparql.rs
```diff
@@ -0,0 +1,186 @@
...
+            value_formats: self
+                .value_formats
+                .clone()
+                .unwrap_or(DsvValueFormats::default(arity)),
+            limit: self.limit,
+            ignore_headers: self.ignore_headers,
```
mmarx, line 179: just hardcode these

mmarx (changes requested):
> how precise should the validation of parameters be? In sparql::is_value_valid() the SparqlParameter::Endpoint and SparqlParameter::Query are validated intensely, which seems like a lot of effort.

Even if this takes a bit of effort, we're doing this once, when loading the program, so the performance overhead is negligible, especially if we have queries with lots of results. Sending off a query to the endpoint, only to _then_ find out that it fails because there's a syntax error, however, is certainly going to take much longer.
 
> The parsing of a resource (now can be either IRI or string) was moved to format::ResouceSpec::parse_string while before, the two file providers HttpResourceProvider and FileResourceProvider were handling this.

As mentioned, this should go into a `TryFrom<Resource>` for `ResourceSpec`.

> in format_builder::new_with_tag the created resource: Option<ResourceSpec> can be replaced by a Option<ResourceSpec> returned from any Builder. This allows the SparqlBuilder to return its own ResourceSpec that uses endpoint and query.

As mentioned, this should rather be another method on the trait, with a default implementation that does not replace the `ResourceSpec`.

> in resource_providers::open_resource() the existing vector containing both ResourceProviders is used. -> can this be simplified somehow?

This should stay as it is. The fact that there's only two entries is an implementation detail that we can't rely on (and might not be true in, e.g., some test cases).

nemo/src/io/formats/sparql.rs
Laschoking: I agree, that the `delimiter` & `ignore_headers` field can be left out here, as mentioned below. The `value_formats`, `limit` seem necessary for the DsvHandler.  The field `parsed_query` is used to compute the number of variables, and could be later manipulated if we want to optimize the query.

nemo/src/io/formats/sparql.rs
Laschoking: You are right that whoever wants a limit can and should just write it inside the query.
So there is no need to support the `limit` field or should we still leave it for completeness since the dsv- and rdf-format also support it?

nemo/src/io/resource_providers/http.rs
Laschoking: Ok. I thought that since `url` and `parameters` are already unpacked we should just pass them. But I understand that your argument makes the code more clear.

nemo/src/io/formats/sparql.rs
```diff
@@ -0,0 +1,186 @@
...
+    type Tag = SparqlTag;
+    fn new(
+        _tag: Self::Tag,
+        parameters: &Parameters<SparqlBuilder>,
+        _direction: Direction,
+    ) -> Result<(Self, Option<ResourceSpec>), ValidationErrorKind> {
+        // Copied from DsvBuilder
+        let value_formats = parameters
+            .get_optional(SparqlParameter::Format)
+            .map(|value| DsvValueFormats::try_from(value).unwrap());
```
mmarx, line 121: I missed this earlier, but this should also be an `except()`

nemo/src/io/formats/sparql.rs
mmarx: Well, if it had been `Query::parse(query.as_str(), None).expect("query has already been validated")`, that would have been clear to me without having to keep the whole context in my head.

nemo/src/io/formats/sparql.rs
mmarx: There's two different things here that we need to answer:
 1. should we allow `limit` as an attribute on `sparql` sources?
 2. do we need to have `limit` as an additional value here in the `SparqlBuilder`.

Certainly the answer to 2. is false, if anything, we should transparently rewrite the query to respect the `limit` attribute; retrieving all the results and then *ignoring all but a few* is just wasteful.

nemo/src/io/formats.rs
```diff
@@ -84,7 +122,7 @@ impl ResourceSpec {
 impl fmt::Display for ResourceSpec {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            ResourceSpec::Resource(r) => f.write_str(r),
+            ResourceSpec::Resource(resource) => f.write_str(resource.to_string().as_str()),
```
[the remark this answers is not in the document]
Laschoking: I wasn't aware of this syntax. however it would be
  `write!(f, "{}", resource.to_string())` right?

nemo/src/io/formats/sparql.rs
mkroetzsch: If the query has already been parsed before, then why do we have to do this again instead of keeping the result from the first time we did it?

nemo/src/io/formats/sparql.rs
```diff
@@ -0,0 +1,186 @@
...
+                .to_iri()
+                .ok_or(ValidationErrorKind::ImportExportInvalidIri)
+                .and_then(|iri| {
+                    Iri::parse(iri)
+                        .map(|_| ())
+                        .map_err(|_| ValidationErrorKind::ImportExportInvalidIri)
+                }),
+            SparqlParameter::Query => value
+                .to_plain_string()
+                .ok_or(ValidationErrorKind::ImportExportLimitNegative)
```
mkroetzsch, line 84: `ImportExportLimitNegative` seems to be incorrect for this case.

Laschoking: That's true. Since it seems  common to call`to_plain_string_unchecked()` instead of `to_plain_string()` the code could also be reduced to
```
Query::parse(&value.to_plain_string_unchecked().as_str(), None)
                    .map(|_| ())
                    .map_err(|e| ValidationErrorKind::ImportExportInvalidSparqlQuery {
                        oxi_error: e.to_string(),
                    })
```

nemo/src/io/format_builder.rs
```diff
@@ -243,14 +243,16 @@ pub(crate) trait FormatBuilder: Sized + Into<AnyImportExportBuilder> {
...
-    ) -> Result<(Self, Option<ResourceSpec>), ValidationErrorKind>;
+    ) -> Result<Self, ValidationErrorKind>;
 
     fn expected_arity(&self) -> Option<usize>;
 
     fn supports_tag(tag: &str) -> bool {
         Self::Tag::from_str(tag).is_ok()
     }
-
+    fn override_resource_spec(&self, _direction: Direction) -> Option<ResourceSpec> {
```
mmarx, line 253: Please leave a blank line between methods

nemo/src/io/formats/sparql.rs
```diff
@@ -114,52 +99,53 @@ impl FormatBuilder for SparqlBuilder {
...
-        ))
+        Ok(Self {
+            value_formats,
+            limit,
+            endpoint,
+            query,
+        })
+    }
+
+    // TODO: do we need to parse parameters here again? that would be annoying
```
mmarx, line 134: We definitely shouldn't need to. All that should be needed here is to replace the IRI with one that also contains the query.

nemo/src/io/formats/sparql.rs
Laschoking: I agree. I think for now we could also not support the `limit` keyword and add support later-on. This could be supported when we try to optimize the query

nemo/src/io/compression_format.rs
```diff
@@ -56,13 +56,13 @@ const GZIP_COMPRESSION_LEVEL: Compression = Compression::new(6);
...
-            resource if resource.ends_with(".gz") => (
+    pub fn from_resource(string: &String) -> (CompressionFormat, String) {
+        match string {
+            string if string.ends_with(".gz") => (
                 CompressionFormat::GZip,
-                resource.as_str()[0..resource.len() - 3].to_string(),
+                string.as_str()[0..string.len() - 3].to_string(),
             ),
-            _ => (CompressionFormat::None, resource.to_owned()),
+            _ => (CompressionFormat::None, string.to_owned()),
```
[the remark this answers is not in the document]
mmarx: Ah, good point. Maybe `from_resource_string`? Still, guessing the correct format is more difficult than just checking if it ends with `.gz` (for IRI's, we'd need to look at the last three octets before fragment and query string)

nemo/src/io/resource_providers/http.rs
```diff
@@ -49,12 +49,7 @@ impl Read for HttpResource {
 }
 
 impl HttpResourceProvider {
-    async fn get(
-        resource: &Resource,
-        url: String,
-        parameters: &HttpParameters,
-        media_type: &str,
-    ) -> Result<HttpResource, ReadingError> {
+    async fn get(resource: &Resource, media_type: &str) -> Result<HttpResource, ReadingError> {
```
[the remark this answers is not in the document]
Laschoking: I went for  `HttpResourceProvider::fetch` since the differences between a GET or POST request are minor in comparison to the shared code

nemo/src/io/compression_format.rs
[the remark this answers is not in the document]
Laschoking: To use the functionality of `Resource::file_extension()` and `Resource::strip_file_extension_unchecked` (maybe a little bit too long name?) we create a temporary `Resource` in `RdfHandler::new`. However it could also be an option that the handlers (`RdfHandler`, `DsvHandler` etc.) have access to the current ResourceBuilder

nemo-physical/src/resource.rs
```diff
@@ -35,17 +69,99 @@ pub enum Resource {
...
+            Self::Iri { parameters, .. } => parameters.post.as_ref(),
+            Self::Path(..) => None,
+        }
+    }
+
+    /// Returns the file extension of a path or Iri, based on the last '.'
+    pub fn file_extension(&self) -> Option<&str> {
+        match self {
+            Self::Path(path) => path.rfind('.').map(|index| &path[index..]),
+            Self::Iri { iri, .. } => iri.authority()?.rfind('.').map(|index| &iri[index..]),
```
mmarx, line 154: `authority` seems to be the wrong part of the IRI here, for `https://example.org/path/to/file.gz`, this would get you `org` as the extension.

nemo-physical/src/resource.rs
```diff
@@ -59,3 +175,112 @@ impl fmt::Display for Resource {
...
+
+/// Builder for a WebResource
+#[derive(Debug, Default)]
+pub struct ResourceBuilder {
+    path: Option<String>,
+    iri: Option<Iri<String>>,
+    headers: Option<Vec<(String, String)>>,
+    get: Option<Vec<(String, String)>>,
+    fragment: Option<String>,
+    post: Option<Vec<(String, String)>>,
```
mmarx, line 187: again, these shouldn't be options

nemo-physical/src/resource.rs
```diff
@@ -59,3 +175,112 @@ impl fmt::Display for Resource {
         }
     }
 }
+
+/// Builder for a WebResource
+#[derive(Debug, Default)]
+pub struct ResourceBuilder {
+    path: Option<String>,
+    iri: Option<Iri<String>>,
```
mmarx, line 183: it's probably cleaner to make this an enum and expose two constructors, so that you can properly error if someone tries to set a header on a path, e.g.

nemo/src/io/compression_format.rs
```diff
@@ -56,13 +56,16 @@ const GZIP_COMPRESSION_LEVEL: Compression = Compression::new(6);
 impl CompressionFormat {
     /// Derive a compression format from the file extension of the given resource,
     /// and return the compression format and the resource string without this extenions.
-    pub fn from_resource(string: &String) -> (CompressionFormat, String) {
-        match string {
-            string if string.ends_with(".gz") => (
+    pub fn from_resource(resource: &Resource) -> (CompressionFormat, String) {
+        let extension = resource.file_extension().unwrap_or("");
+
+        if extension == ".gz" {
```
mmarx, line 62: `if resource.file_extension() == Some(".gz")`?

nemo/src/io/compression_format.rs
```diff
@@ -56,13 +56,16 @@ const GZIP_COMPRESSION_LEVEL: Compression = Compression::new(6);
...
+            (
                 CompressionFormat::GZip,
-                string.as_str()[0..string.len() - 3].to_string(),
-            ),
-            _ => (CompressionFormat::None, string.to_owned()),
+                resource.strip_file_extension_unchecked(".gz").to_string(), //string.as_str()[0..string.len() - 3].to_string(),
```
mmarx, line 65: please remove the comment

nemo/src/io/format_builder.rs
```diff
@@ -210,6 +213,10 @@ format_parameter! {
     pub(crate) enum StandardParameter(NoParameters) {
         Resource(name = attribute::RESOURCE, supported_types = &[ValueType::String, ValueType::Constant]),
         Compression(name = attribute::COMPRESSION, supported_types = &[ValueType::String]),
+        //HttpHeader(name = attribute::HTTP_HEADERS, supported_types= &[ValueType::String]),
+        //HttpHeader(name = attribute::HTTP_GET, supported_types= &[ValueType::String]),
+        //HttpHeader(name = attribute::HTTP_FRAGMENT, supported_types= &[ValueType::String]),
+        //HttpHeader(name = attribute::HTTP_POST, supported_types= &[ValueType::String]),
```
mmarx, line 219: What's this?

nemo/src/io/format_builder.rs
```diff
@@ -250,9 +257,11 @@ pub(crate) trait FormatBuilder: Sized + Into<AnyImportExportBuilder> {
     fn supports_tag(tag: &str) -> bool {
         Self::Tag::from_str(tag).is_ok()
     }
-    fn override_resource_spec(&self, _direction: Direction) -> Option<ResourceSpec> {
+
+    fn override_resource_builder(&self, _direction: Direction) -> Option<ResourceBuilder> {
         None
     }
+
```
mmarx, line 264: if you're still passing around the `ResourceBuilder` at this stage, this should rather look like 
```rust
fn customize_resource_builder(&self, _direction: Direction, builder: ResourceBuilder) -> ResourceBuilder {
  builder
}
```

nemo/src/io/format_builder.rs
```diff
@@ -397,21 +406,17 @@ impl ImportExportBuilder {
...
-                            Err(kind) => {
-                                // Report the error and return None
-                                builder.report_error(origin, kind);
-                                None
-                            }
-                        }
+            .and_then(|value| {
+                ResourceBuilder::try_from(value)
+                    .map_err(|err| {
+                        // Error conversion should work since ResourceValidationErrorKind is registered in ValidataionErrorKind
```
mmarx, line 415: What is this comment doing? If this wasn't the case, the code wouldn't compile. What information does it add?

nemo/src/io/formats/sparql.rs
```diff
@@ -125,21 +123,23 @@ impl FormatBuilder for SparqlBuilder {
...
-                self.query.to_sse(),
-            ))
+            let mut resource_builder = ResourceBuilder::from_valid_iri(self.endpoint.clone());
+            let query = self.query.to_string();
+            if query.len() > HTTP_GET_CHAR_LIMIT {
+                resource_builder.parameter_mut(HttpParamType::Post, String::from("query"), query);
+            } else {
+                resource_builder.parameter_mut(HttpParamType::Get, String::from("query"), query);
+            }
+            Some(resource_builder)
```
mmarx, line 142: Would be cleaner (and safer, should `Direction` ever get another variant) to have a `match direction { … }` here

nemo/src/io/resource_providers/http.rs
```diff
@@ -87,38 +97,32 @@ impl HttpResourceProvider {
...
-            let enc_query = encode(query);
+        let response = {
+            // Collect Body-Parameters into one string
+            if let Some(body) = resource.body() {
+                // TODO: Verify if RequestBuilder::form() can handle the body correctly
+                //let body_string = body
+                //    .iter()
+                //    .map(|(key, value)| format!("{}={}", key, value))
+                //    .collect::<Vec<String>>()
+                //    .join("&");
```
mmarx, line 110: can this go?

nemo/src/io/resource_providers/http.rs
```diff
@@ -144,23 +148,15 @@ impl ResourceProvider for HttpResourceProvider {
...
-            return Ok(None);
-        }
 
         let rt = tokio::runtime::Builder::new_current_thread()
             .enable_all()
             .build()
             .map_err(|e| ReadingError::from(e).with_resource(resource.clone()))?;
 
-        let response = rt.block_on(Self::get(resource, media_type))?;
+        // Verify if resource has a non-empty body
```
mmarx, line 157: (i) do you mean `response` instead of `resource`?
(ii) why would it need to have a non-empty body?

nemo/src/rule_model/components/import_export/attributes.rs
```diff
@@ -53,11 +53,31 @@ pub enum ImportExportAttribute {
     #[assoc(from_name = attribute::ENDPOINT)]
     #[assoc(value_type = ProgramComponentKind::Iri)]
     Endpoint,
-    /// The query that is evaluated by the endpoint
+    /// The query send to the endpoint
```
mmarx, line 56: `sent`

nemo/src/rule_model/error/validation_error.rs
```diff
@@ -167,9 +168,13 @@ pub enum ValidationErrorKind {
     #[assoc(code = 233)]
     ImportExportInvalidIri,
     /// Import/Export: invalid SPARQL query
-    #[error(r#"invalid Sparql query resulting in oxigraph-parser error: `{oxi_error}`"#)]
     #[assoc(code = 234)]
+    #[error(r#"invalid Sparql query resulting in oxigraph-parser error: `{oxi_error}`"#)]
```
mmarx, line 172: error messages shouldn't be unnecessarily verbose, this should be something like 
```rust
"invalid SPARQL query: `{oxi_error}`"
```

nemo/src/syntax.rs
```diff
@@ -403,8 +403,16 @@ pub mod import_export {
         pub const IGNORE_HEADERS: &str = "ignore_headers";
         /// The SPARQL endpoint for the query
         pub const ENDPOINT: &str = "endpoint";
-        /// The SPARQL query that is appended to the GET-request.
+        /// The SPARQL query that send to the endpoint.
```
mmarx, line 406: `The SPARQL query that is sent to the endpoint.`

nemo/src/syntax.rs
```diff
@@ -403,8 +403,16 @@ pub mod import_export {
...
         pub const QUERY: &str = "query";
+        /// Name of Parameters that are passed as HTTP-headers.
+        pub const HTTP_HEADERS: &str = "http_headers";
+        /// Name of Parameters that are appended to a web IRI.
+        pub const HTTP_GET: &str = "http_get_parameters";
```
mmarx, line 411: for consistency, this should be `HTTP_GET_PARAMETERS`

nemo/src/syntax.rs
```diff
@@ -403,8 +403,16 @@ pub mod import_export {
...
+        /// Name of the attribute that is appended to the web IRI.
+        pub const HTTP_FRAGMENT: &str = "http_fragment";
+        /// Name of Parameters that are send in the body of a POST request.
+        pub const HTTP_POST: &str = "http_post_parameters";
```
mmarx, line 415: for consistency, this should be `HTTP_POST_PARAMETERS`

nemo/src/rule_model/components/import_export/attributes.rs
```diff
@@ -53,11 +53,31 @@ pub enum ImportExportAttribute {
...
+    /// The parameters appended to the web-IRI
+    #[assoc(name = attribute::HTTP_GET)]
+    #[assoc(from_name = attribute::HTTP_GET)]
+    #[assoc(value_type = ProgramComponentKind::PlainString)]
+    HttpGet,
+    /// The parameters send in the body of a POST-request
+    #[assoc(name = attribute::HTTP_POST)]
+    #[assoc(from_name = attribute::HTTP_POST)]
+    #[assoc(value_type = ProgramComponentKind::PlainString)]
+    HttpPost,
```
mmarx, line 70: These should be plural as well (see below)

nemo/src/io/format_builder.rs
```diff
@@ -429,12 +434,25 @@ impl ImportExportBuilder {
...
+
+        let resource_spec = match resource_builder {
+            Some(mut rb) => {
+                if rb.has_iri() {
+                    // TODO: read Header, Get, Fragment & POST and add them to rb
+                }
+                Some(ResourceSpec::Resource(rb.finalize()))
+            }
+            None => None,
+        };
```
mmarx, line 449: this can be 
```rust
let resource_spec = resource_builder.map(|builder| {
  // …
  ResourceSpec::Resource(builder.finalize())
});
```

nemo/src/io/format_builder.rs
```diff
@@ -388,9 +394,27 @@ impl ImportExportBuilder {
...
-        let resource = parameters
+        // Convert resource from type String or Iri into a ResourceSpec
+        let resource: Option<ResourceSpec> = parameters
             .get_optional(StandardParameter::Resource.into())
-            .map(|resource| ResourceSpec::from_string(resource.to_plain_string_unchecked()));
+            .and_then(|path| {
+                path.to_plain_string()
+                    .or_else(|| path.to_iri())
+                    .and_then(|string| {
+                        match ResourceSpec::parse_string(string) {
```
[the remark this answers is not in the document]
Laschoking: with the `ResourceBuilder` we now use `TryFrom<AnyDataValue>`

nemo-physical/src/resource.rs
```diff
@@ -35,17 +69,99 @@ pub enum Resource {
...
+    /// Return the headers if resource is of variant [Resource::Iri]
+    pub fn headers(&self) -> Option<&Vec<(String, String)>> {
+        match self {
+            Self::Iri { parameters, .. } => parameters.headers.as_ref(),
+            Self::Path(..) => None,
+        }
+    }
+
+    /// Return the body parameter of HttpParameters
+    pub fn body(&self) -> Option<&Vec<(String, String)>> {
```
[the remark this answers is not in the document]
Laschoking: changed the function name. There were no `get_parameters` because I did not see any use case for this function, since the final web-Iri will contain the get-parameters already. I am adding  the function for completeness though.

nemo/src/io/format_builder.rs
Laschoking: this was commented since the reading and validation of the Http-parameters  is not yet implemented. Once it is implemented, the code above (with correct formulation) will be included to handle Http-Parameters as part of the `StandardParameter`

nemo/src/io/resource_providers/http.rs
Laschoking: this comment must have slipped my attention during code refactoring.  In an outdated version it was verified in this snipped whether a `resource` had a body and then would decide for a POST-request or otherwise for a GET-request.

nemo-physical/src/resource.rs
```diff
@@ -35,17 +69,99 @@ pub enum Resource {
...
+        }
+    }
+
+    /// Remove a certain file extension from the resource
+    pub fn strip_file_extension_unchecked(&self, suffix: &str) -> &str {
+        match self {
+            Self::Path(path) => path
+                .strip_suffix(suffix)
+                .expect("suffix should ve verified"),
+            Self::Iri { iri, .. } => iri.strip_suffix(suffix).expect("suffix should ve verified"),
```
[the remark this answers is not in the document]
Laschoking: `be`

nemo-physical/src/resource.rs
Laschoking: I replace the lines with `location: ResourceLocation`  for
```
pub enum ResourceLocation {
    Path(String),
    Iri(Iri<String>),
}
```

nemo/src/io/format_builder.rs
Laschoking: You're right, I will remove the comment.

nemo/src/rule_model/components/import_export/attributes.rs
Laschoking: i will note this for a later issue/PR and stick with `PlainString` for now

nemo/src/io/formats.rs
```diff
@@ -57,15 +62,48 @@ pub enum ResourceSpec {
...
-            Self::Resource(string)
+            Self::Resource(Resource::Path(string))
         }
     }
 
+    /// Parse and valiate [String] to a [ResourceSpec].
+    pub(crate) fn parse_string(string: String) -> Result<Self, ValidationErrorKind> {
+        match string {
+            // Convert the web-resource into a valid Iri
+            string if string.starts_with("http:") || string.starts_with("https:") => {
```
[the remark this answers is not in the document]
Laschoking: The code is moved to `ResourceBuilder::try_from`. I agree, it is more simple to use the parsed IRI to check the scheme. However if we also want to strip a local Iri of `file://localhost` this cannot be done by simply removing the scheme (since `localhost` is not part of it)

nemo-physical/src/resource.rs
Laschoking: fair, i'll replace `authority` with `path`

nemo-physical/src/resource.rs
```diff
@@ -1,10 +1,61 @@
...
+        /// Contains the endpoint / http address for a request in the form of an Iri
+        iri: Iri<String>,
+        /// A collection of parameters that are passed considered in the http-request
+        /// Each tuple consists of a keyword & the value e.g. ("query", "SELECT ...")
+        /// Currently query is the onliest keyword allowed
+        parameters: HttpParameters,
+    },
+}
+
+impl Resource {
```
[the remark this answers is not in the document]
Laschoking: I added functions to retrieve the parameters as well as `as_path()` and `as_iri()` functions which only work if Resource has the correct variant `Path` or `Iri`. `as_string()` is can be used if we dont care about the type of the resource.

nemo-physical/src/resource.rs
```diff
@@ -1,10 +1,502 @@
...
+        if self.headers.iter().any(|(k, _)| k == &key) {
+            Err(ResourceValidationErrorKind::ImportExportDuplicateHttpHeader(key))
+        } else {
+            self.headers.push((key, value));
+            Ok(self)
+        }
+    }
+
+    /// Add key-value pair to GetParameter
+    pub fn get_mut(&mut self, key: String, value: String) -> &Self {
```
mmarx, line 401: * `should be called `add_query_parameter` or `add_get_parameter`
* should return `&mut Self`, so that calls can be chained

nemo-physical/src/resource.rs
```diff
@@ -1,10 +1,502 @@
...
+    ) -> Result<&mut Self, ResourceValidationErrorKind> {
+        if self.fragment.is_some() {
+            return Err(ResourceValidationErrorKind::ImportExportDuplicateFragment);
+        } else {
+            self.fragment = Some(value);
+        }
+        Ok(self)
+    }
+
+    /// Add key-value pair to PostParameter
```
mmarx, line 419: `Add POST parameter`

nemo-physical/src/resource.rs
```diff
@@ -1,10 +1,502 @@
...
+    pub fn post_mut(&mut self, key: String, value: String) -> &Self {
```
mmarx, line 420: * `add_post_parameter`
* should return `&mut Self`

nemo-physical/src/resource.rs
```diff
@@ -1,10 +1,502 @@
...
+        self.post_parameters.push((key, value));
+        self
+    }
+
+    /// Builds a [Resource] with type [Resource::Iri] or [Resource::Path] that contains all registered parameters
```
mmarx, line 425: `Build a [Resource] with the given parameters.`

nemo-physical/src/resource.rs
```diff
@@ -1,10 +1,502 @@
...
+    pub fn finalize(self) -> Result<Resource, ResourceValidationErrorKind> {
```
mmarx, line 426: (see below)
```rust
pub fn finalize(self) -> Resource {
  self.resource
}
```

nemo/src/io/format_builder.rs
```diff
@@ -209,6 +213,10 @@ format_parameter! {
     pub(crate) enum StandardParameter(NoParameters) {
         Resource(name = attribute::RESOURCE, supported_types = &[ValueType::String, ValueType::Constant]),
         Compression(name = attribute::COMPRESSION, supported_types = &[ValueType::String]),
+        HttpHeaders(name = attribute::HTTP_HEADERS, supported_types= &[ValueType::Map]),
+        HttpGetParameters(name = attribute::HTTP_GET_PARAMETERS, supported_types= &[ValueType::Map]),
+        HttpFragment(name = attribute::HTTP_FRAGMENT, supported_types= &[ValueType::Map]),
```
mmarx, line 218: see elsewhere (not sure whether that was above or below, I've lost track …), this shouldn't have `HTTP` in its name, and probably should not even be an attribute.

nemo/src/io/format_builder.rs
```diff
@@ -407,9 +441,77 @@ impl ImportExportBuilder {
...
         }?;
 
+        let resource_builder = inner
+            .override_resource_builder(direction)
+            .or(resource_builder);
+
+        let resource = {
+            if let Some(mut rb) = resource_builder {
+                // Unpack Http-Parameters and add them to the ResourceBuilder
+                // Header-Parameters
```
mmarx, line 451: This comment can go, it's not adding anything.

nemo/src/io/format_builder.rs
```diff
@@ -407,9 +441,77 @@ impl ImportExportBuilder {
...
+                let headers = parameters
+                    .get_optional(StandardParameter::HttpGetParameters.into())
+                    .map(ResourceBuilder::unpack_header_parameter);
+
+                if let Some(iter) = headers {
```
mmarx, line 456:
```rust
if let Some(headers) = parameters.get_optional(…).map(…) {
  for (key, value) in headers { … }
}
```

nemo/src/io/format_builder.rs
```diff
@@ -407,9 +441,77 @@ impl ImportExportBuilder {
...
+                    for (key, value) in iter {
+                        rb.header_mut(key, value)
+                            .map_err(|err| builder.report_error(origin, err.into()))
+                            .ok()?;
+                    }
+                }
+
+                // GET-Parameters
```
mmarx, line 464: At least get rid of the hyphen and downcase parameters, better get rid of the comment.

nemo/src/io/format_builder.rs
```diff
@@ -407,9 +441,77 @@ impl ImportExportBuilder {
...
+                let get_parameters = parameters
+                    .get_optional(StandardParameter::HttpGetParameters.into())
+                    .map(ResourceBuilder::unpack_get_post_parameter);
+
+                if let Some(iter) = get_parameters {
```
mmarx, line 469: see above, move the `if let` outside

nemo/src/io/format_builder.rs
```diff
@@ -407,9 +441,77 @@ impl ImportExportBuilder {
...
+                    for (key, value) in iter {
+                        rb.get_mut(key, value);
+                    }
+                }
+
+                // POST-Parameters
```
mmarx, line 475: see above

nemo/src/io/format_builder.rs
```diff
@@ -407,9 +441,77 @@ impl ImportExportBuilder {
...
+                let post_parameters = parameters
+                    .get_optional(StandardParameter::HttpGetParameters.into())
+                    .map(ResourceBuilder::unpack_get_post_parameter);
+
+                if let Some(iter) = post_parameters {
```
mmarx, line 480: see above

nemo/src/io/format_builder.rs
```diff
@@ -407,9 +441,77 @@ impl ImportExportBuilder {
...
+                if let Some(fragment) = parameters
+                    .get_optional(StandardParameter::HttpFragment.into())
+                    .map(|value| value.lexical_value())
+                {
+                    rb.fragment_mut(fragment)
+                        .map_err(|err| builder.report_error(origin, err.into()))
+                        .ok()?;
+                }
+
+                // Report a possible ResourceValidationErrorKind and propagate possible failure upwards
```
mmarx, line 496: Just repeats what the code already says.

nemo/src/io/format_builder.rs
```diff
@@ -407,9 +441,77 @@ impl ImportExportBuilder {
             }
         }?;
 
+        let resource_builder = inner
+            .override_resource_builder(direction)
+            .or(resource_builder);
```
mmarx, line 446: Why not make this `unwrap_or(resource_builder);` and get rid of an `if let` with an `else` that is _way_ too far along, below?

nemo/src/io/format_builder.rs
```diff
@@ -407,9 +441,77 @@ impl ImportExportBuilder {
...
+                        builder.report_error(origin, err.into())
+                    })
+                    .ok()?;
+                Some(resource)
+            } else {
+                None
+            }
+        };
+
+        // Build the Resource and ResourceSpec once ResourceBuilder collected all information
```
mmarx, line 509: Adds nothing, get rid of it.

nemo/src/io/format_builder.rs
```diff
@@ -407,9 +441,77 @@ impl ImportExportBuilder {
...
+        let resource_spec = resource.map(ResourceSpec::Resource);
+
         Some(ImportExportBuilder {
             inner: inner.into(),
-            resource,
+            resource: resource_spec,
             compression,
         })
```
mmarx, line 516:
```rust
Some(ImportExportBuilder {
            inner: inner.into(),
            resource: resource.map(ResourceSpec::Resource),
            compression,
        })
```

nemo/src/io/formats/sparql.rs
```diff
@@ -0,0 +1,172 @@
...
+                }
+            }),
+            SparqlParameter::Endpoint => value
+                .to_iri()
+                .ok_or(ValidationErrorKind::ImportExportInvalidIri)
+                .and_then(|iri| {
+                    Iri::parse(iri)
+                        .map(|_| ())
+                        .map_err(|_| ValidationErrorKind::ImportExportInvalidIri)
+                }),
```
mmarx, line 71: This shouldn't need to throw the same error twice, clearly, `to_iri()` should actually return a validated IRI.

mmarx (changes requested):
This is going in the right direction. I have a lot of comments, but most of them are stylistic issues (particularly in user-facing code such as error messages).

You can ignore the clippy issues for now and rebase once #626 is merged, they're already fixed there. The doc failure is already fixed, next run should work again.

nemo/src/io/format_builder.rs
```diff
@@ -489,6 +596,7 @@ pub(crate) enum AnyImportExportBuilder {
     Dsv(DsvBuilder),
     Rdf(RdfHandler),
     Json(JsonHandler),
+    Sparql(SparqlBuilder),
```
mmarx, line 599: This enum needs an `#[allow(variant_size_differences)]` (but GitHub won't let me annotate the enum itself, since it wasn't changed)

nemo-physical/src/resource.rs
```diff
@@ -1,10 +1,502 @@
...
+#[derive(Clone, Debug, Error)]
+pub enum ResourceValidationErrorKind {
+    /// Resource is not a valid path
+    #[error("Invalid Path")]
+    ImportExportInvalidPath,
+    /// Parsing error for Iri
+    #[error("Invalid IRI format")]
+    ImportExportInvalidIri,
+    /// Http-Parameters have wrong type
+    #[error("Invalid Http-parameter: {0}")]
```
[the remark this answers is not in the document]
Laschoking: for now `{0}` contains a specification which element of the parameter has unexpected type.

nemo-physical/src/resource.rs
```diff
@@ -1,10 +1,502 @@
...
+
+    /// Return iri of a resource
+    pub fn as_iri(&self) -> String {
+        Self::internal_as_iri(self, false)
+    }
+
+    /// Return iri of a resource
+    pub fn as_iri_encoded(&self) -> String {
+        Self::internal_as_iri(self, true)
+    }
```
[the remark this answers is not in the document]
Laschoking: sure. I just thought for debugging purposes it would be nicer to have a non-encoded function

nemo/src/io/compression_format.rs
```diff
@@ -56,13 +56,14 @@ const GZIP_COMPRESSION_LEVEL: Compression = Compression::new(6);
 impl CompressionFormat {
     /// Derive a compression format from the file extension of the given resource,
     /// and return the compression format and the resource string without this extenions.
-    pub fn from_resource(resource: &Resource) -> (CompressionFormat, Resource) {
-        match resource {
-            resource if resource.ends_with(".gz") => (
+    pub fn from_resource(resource: &Resource) -> (CompressionFormat, String) {
```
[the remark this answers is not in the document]
Laschoking: Before the latest changes `Resource` was a type alias to `String`.
The only use case of this function is in the `RdfHandler`  which expects a String-like value so I thought we want to return a String.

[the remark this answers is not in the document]
mmarx: it then immediately goes on to turn that `String` back into a `Path`, just so that it can call `extensions()` on it. If this returned `Resource`, we could skip those conversions and the code would be simpler.

nemo/src/syntax.rs
```diff
@@ -399,10 +399,22 @@ pub mod import_export {
...
+        /// The SPARQL endpoint for the query
+        pub const ENDPOINT: &str = "endpoint";
+        /// The SPARQL query that is send to the endpoint.
+        pub const QUERY: &str = "query";
+        /// Name of Parameters that are passed as HTTP-headers.
+        pub const HTTP_HEADERS: &str = "http_headers";
+        /// Name of Parameters that are appended to a web IRI.
+        pub const HTTP_GET_PARAMETERS: &str = "http_get_parameters";
+        /// Name of the attribute that is appended to the web IRI.
+        pub const HTTP_FRAGMENT: &str = "http_fragment";
```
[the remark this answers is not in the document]
mmarx: `IRI_FRAGMENT` is certainly the better name, yeah. Let's leave it in for now, we can always rip it out again later if we decide against it (I'm leaning slightly towards there being only one way to specify this, not two, and making it illegal to specify a fragment as part of the IRI feels wrong, though).

nemo/src/io/format_builder.rs
Laschoking: I replaced it with `IriFragment` until we decide if we want to keep it as a separate parameter

nemo-physical/src/resource.rs
```diff
@@ -1,10 +1,502 @@
...
+            } else {
+                get_parameters.push('&');
+            }
+            // TODO: could also be replaced with an inline function
+            if encode_get {
+                get_parameters.push_str(&format!("{}={}", encode(key), encode(value)));
+            } else {
+                get_parameters.push_str(&format!("{}={}", key, value));
+            }
+        }
```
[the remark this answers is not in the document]
Laschoking: the solution  with `serde_urlencoding` looks more intuitive but runs into a Serialization error:
"the trait bound `dyn Iterator<Item = &(std::string::String, std::string::String)>: Serialize` is not satisfied".
Since we had a similar problem with the `form` parameter of of an HTTP POST request, I would stick with the first solution for now.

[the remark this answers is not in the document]
Laschoking: yes I didn't see this shortcut. That solves the matter.

nemo-physical/src/resource.rs
```diff
@@ -1,10 +1,502 @@
...
+        match iri.authority() {
+            None | Some("") | Some("localhost") => Ok(String::from(iri.path())),
+            _ => Err(ResourceValidationErrorKind::ImportExportInvalidIri),
+        }
+    }
+
+    /// Verify if [ResourceBuilder] contains a valid web-Iri
+    pub fn has_web_iri(&self) -> bool {
+        matches!(self.location, ResourceLocation::Iri(..))
+    }
```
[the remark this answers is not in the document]
Laschoking: I'll remove the function as it does not have any use any more.

nemo-physical/src/resource.rs
```diff
@@ -1,10 +1,502 @@
...
+
+    /// Validate key-value pairs of HttpHeaders
+    pub fn validate_header(map: AnyDataValue) -> Result<(), ResourceValidationErrorKind> {
```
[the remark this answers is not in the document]
Laschoking: As you pointed out, the function should be called `validate_headers` .

nemo-physical/src/resource.rs
```diff
@@ -1,10 +1,502 @@
...
+        matches!(self, Self::Path(..))
+    }
+
+    /// Return local path of a resource
+    pub fn as_path(&self) -> &PathBuf {
+        match self {
+            Self::Iri { .. } => panic!("Resource no local path"),
+            Self::Path(path) => path,
+        }
+    }
```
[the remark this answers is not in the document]
Laschoking: this function is now included through the Display implementation.

nemo-physical/src/resource.rs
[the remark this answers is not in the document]
mmarx: > The validation functions can also be separated from the ResourceBuilder if this is more convenient.

No, the point of the `ResourceBuilder` is to make it impossible to build invalid `Resources` without going through `ResourceBuilder`, which enforces that all parameters are valid (that is also why `Resource` should not have a public constructor). But then there's no need for `validate_header`, since the `header` method for adding a new header should already do the validation.

nemo-physical/src/resource.rs
```diff
@@ -1,10 +1,502 @@
...
+        let keys = map.map_keys();
+
+        let mut vec: Vec<(String, String)> = Vec::new();
+        if let Some(keys_iter) = keys {
+            for key in keys_iter {
+                let element = map.map_element_unchecked(key);
+                vec.push((key.lexical_value(), element.lexical_value()));
+            }
+        }
+        vec.into_iter()
```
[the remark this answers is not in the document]
mmarx: the `flat_map` is there to deal with the `Option` returned from `map_keys`. You're right, though, that this should be
```rust
let result = keys.into_iter().flat_map(|keys| keys.map(|key| (key.lexical_value(), map.map_element_unchecked(key).lexical_value()))).collect::<Vec<_>>();
        result.into_iter()
```

nemo-physical/src/resource.rs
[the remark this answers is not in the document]
Laschoking: I was under the impression that as much as possible validation should already happen in the `is_value_valid` function in `format_builder`. However this function looks rather like a quick check than an in-depth validation. 
Following your opinion we would do the entire validation of the HTTP parameters and fragment once they are added to the `ResourceBuilder`, right?

nemo-physical/src/resource.rs
```diff
@@ -1,10 +1,502 @@
...
+        match Iri::parse(string.to_owned()) {
+            Ok(iri) => {
+                let scheme = iri.scheme();
+                debug!("scheme:{}", scheme);
+                match scheme {
+                    "http" | "https" => Ok(Self::from(iri)),
+                    "file" => Ok(Self::from(Self::strip_local_iri(iri)?)),
+                    _ => Err(ResourceValidationErrorKind::ImportExportUnknownIriScheme(
+                        String::from(scheme),
+                    )),
```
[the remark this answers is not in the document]
Laschoking: `value.to_iri()` returns `Option<String>` so with you suggestion we would always end up in the `TryFrom<String>` case except there is a ValueDomain error.
I guess the IRI parsing of the string can happen in `TryFrom<String>` and then call `From<Iri<String>>` if the string is an IRI.

nemo-physical/src/resource.rs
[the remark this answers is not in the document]
mmarx: This might be conflating two different concerns:
* the `ResourceBuilder` should assure that there is no way to construct invalid resources, anywhere
* `is_value_valid` is for user-facing error reporting

The natural way to reconcile these two is to use the `ResourceBuilder` in `is_value_valid` to construct the resource.

nemo-physical/src/resource.rs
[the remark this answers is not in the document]
mmarx: Yes, that's precisely what I meant: implement `TryFrom<String>` (which should try to parse this as an IRI, matching on the scheme to decide whether it's a path or not), and then just call this.

nemo/src/io/formats.rs
```diff
@@ -57,12 +59,12 @@ pub enum ResourceSpec {
...
 
 impl ResourceSpec {
-    /// Convert a [String] to a [ImportExportResource].
+    /// Convert a simple [String] into a [ResourceSpec]
     pub(crate) fn from_string(string: String) -> Self {
         if string.is_empty() {
             Self::Stdout
         } else {
-            Self::Resource(string)
+            Self::Resource(Resource::Path(PathBuf::from(string)))
```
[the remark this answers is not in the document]
Laschoking: I agree. Since it is actually `TryFrom<String>` I was thinking about this:
`Resource(ResourceBuilder::try_from(string).expect("Internally created string is not valid").finalize())` 
Do you agree that the `expect()` is justified?

nemo/src/io/formats/rdf.rs
```diff
@@ -211,7 +212,9 @@ impl FormatBuilder for RdfHandler {
                 let value = parameters
                     .get_required(RdfParameter::BaseParamType(StandardParameter::Resource));
 
-                let resource = value.to_plain_string_unchecked();
+                // Use functionality of ResourceBuilder to read the Resource parameter
+                // and create a dummy resource
```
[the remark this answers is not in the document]
Laschoking: I called it a `dummy` resource, because the resource  is only constructed to detect the extension and not adopted elsewhere

nemo/src/io/formats/rdf.rs
```diff
@@ -211,7 +212,9 @@ impl FormatBuilder for RdfHandler {
                 let value = parameters
                     .get_required(RdfParameter::BaseParamType(StandardParameter::Resource));
 
-                let resource = value.to_plain_string_unchecked();
+                // Use functionality of ResourceBuilder to read the Resource parameter
```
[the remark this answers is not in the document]
Laschoking: I'll remove it.

nemo-wasm/src/lib.rs
```diff
@@ -191,7 +191,11 @@ impl ResourceProvider for BlobResourceProvider {
         resource: &Resource,
         _media_type: &str,
     ) -> Result<Option<Box<dyn std::io::Read>>, nemo_physical::error::ReadingError> {
-        if let Some(blob) = self.blobs.get(resource) {
+        let res = match resource {
+            Resource::Path(path) => path,
+            Resource::Iri { iri, .. } => &iri.to_string()
```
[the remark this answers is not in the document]
Laschoking: can be handled with `resource.to_string()` now

nemo/src/io/formats.rs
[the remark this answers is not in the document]
mmarx: First of all, the message should be “is valid”, not “is *not* valid”. Then, if it can panic, the method needs to be called `from_string_unchecked`.

However, since we only ever use actually this for the fallback file names, better replace it with something like this:
```rust
pub(crate) fn default_resource(predicate_name: &str, meta: &dyn FileFormatMeta) -> Self {
  ResourceBuilder::try_from(format!("{predicate_name}.{}", meta.default_extension()).expect("default name is valid").finalize())
}
```

nemo/src/io/formats/rdf.rs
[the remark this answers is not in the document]
mmarx: So, this is more or less obsolete, since we'd actually want to keep passing a resource around here, not a `String`, but in this case, a comment such as “Create a resource to extract the file extension” would have been helpful: it communicates *why* we do something, not *what* we do (since that's already stated in the code)

nemo/src/io/formats.rs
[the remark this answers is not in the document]
Laschoking: This would however require dynamic upcasting for the handlers of `Arc<dyn ImportHandler + Send + Sync>:` and `Arc<dyn ExportHandler + Send + Sync>:`  to `&dyn FileFormatMeta` which is supported via `#![feature(trait_upcasting)]`.
The other option would be to pass the `default_extension` as a parameter into `default_resource`.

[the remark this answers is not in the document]
mmarx: Perfectly fine to enable it, it's [going to be stable in 1.86](https://releases.rs/docs/1.86.0/), which is due to release in next month.

nemo/src/io/formats.rs
```diff
@@ -57,15 +62,48 @@ pub enum ResourceSpec {
...
+        }
+    }
+
+    /// Convert a [String] to a [ResourceSpec].
+    pub(crate) fn from_endpoint(endpoint: Iri<String>, query: String) -> Self {
+        Self::Resource(Resource::Iri {
+            iri: endpoint,
+            parameters: HttpParameters { query: Some(query) },
+        })
+    }
```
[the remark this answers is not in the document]
Laschoking: This is outdated with the `ResourceBuilder`.

nemo/src/io/resource_providers.rs
```diff
@@ -20,7 +20,11 @@ pub mod http;
 ///
 /// TODO: Revise if deemed necessary.
 fn is_iri(resource: &Resource) -> bool {
-    resource.contains(':')
+    match resource {
+        Resource::Path(path) => path.contains(':'),
```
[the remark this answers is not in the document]
Laschoking: outdated with the Display implementation of a `Resource`

nemo/src/io/resource_providers.rs
```diff
@@ -20,7 +20,11 @@ pub mod http;
...
+        // TODO: We would expect Resource::Iri to be an iri. However is this verified somewhere already?
+        Resource::Iri { iri, .. } => iri.to_string().contains(':'),
+    }
```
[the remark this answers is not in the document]
Laschoking: outdated since we try all available `ResourceProviders`

nemo-physical/src/resource.rs
```diff
@@ -1,10 +1,502 @@
...
+    /// Returns the file extension of a path or Iri, based on the last '.'
+    pub fn file_extension(&self) -> Option<&str> {
+        match self {
+            Self::Path(path) => path.extension().and_then(|ext| ext.to_str()),
+            Self::Iri { iri, .. } => iri.path().rfind('.').map(|index| &iri[index..]),
+        }
+    }
+
+    /// Remove a certain file extension from the resource
+    pub fn strip_file_extension_unchecked(&self, suffix: &str) -> String {
```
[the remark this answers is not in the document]
Laschoking: I agree, it should return the modified resource!

nemo-physical/src/resource.rs
```diff
@@ -1,10 +1,502 @@
...
+    /// Create a new [ResourceBuilder] from a web-Iri
+    fn from(iri: Iri<String>) -> Self {
+        Self {
+            location: ResourceLocation::Iri(iri),
+            headers: Vec::new(),
+            get_parameters: Vec::new(),
+            fragment: None,
+            post_parameters: Vec::new(),
+        }
+    }
```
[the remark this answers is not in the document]
Laschoking: Do you see cases where we want to include the `query` and `fragment` segment of the local IRI into the `Path` or should their existence rather cause an error?

nemo-physical/src/resource.rs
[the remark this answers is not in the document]
Laschoking: Just to be clear, there are no objections if the mentioned scheme matching of the IRI happens in the `TryFrom<Iri<String>>` right?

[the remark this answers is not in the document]
mmarx: That's precisely where the matching should happen, yes. Sorry if that still wasn't totally clear …

nemo/src/io/format_builder.rs
Laschoking: The `resource_builder` from above is of type `Option<ResourceBuilder>` , because  that `resource` parameter is optional. So in the case that neither a `resource` parameter nor the `enpoint` paramter are supplied there is no `resource_builder`.  I guess the `if-else` branch could be avoided via `resource_builder.map(|rb| {... rb.finalize()}`

nemo/src/io/resource_providers/http.rs
```diff
@@ -66,26 +64,66 @@ impl HttpResourceProvider {
...
         let client = reqwest::Client::builder()
             .default_headers(headers)
             .build()
             .map_err(err_mapping)?;
 
-        let response = client.get(url).send().await.map_err(err_mapping)?;
+        let full_url = resource.as_iri_encoded();
+
+        let response = {
+            // Collect Body-Parameters into one string
```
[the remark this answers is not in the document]
Laschoking: removed obsolete comment

nemo/src/io/format_builder.rs
mmarx: yeah, you're right.

nemo-physical/src/resource.rs
```diff
@@ -94,383 +96,349 @@ impl Resource {
...
-        if self.fragment.is_some() {
-            return Err(ResourceValidationErrorKind::ImportExportDuplicateFragment);
-        } else {
-            self.fragment = Some(value);
+        match &mut self.resource {
+            Resource::Http { parameters, .. } => {
+                parameters.post_parameters.push((key, value));
+                Ok(self)
+            }
+            Resource::Path(..) => Err(ResourceValidationErrorKind::UnexpectedHttpParameter),
```
mmarx, line 375: See above, catchall clause

mmarx (changes requested):
This has improved a lot! I've found a few small issues, though:
* the merge reintroduced (unused) dependencies on the rio crates, which we should get rid of
* an error message swapped two parameters
* parts of the validation logic currently in the `ResourceBuilder` shouldn't actually be there, but rather live somewhere in `nemo` (probably in a new module below `nemo::io`, maybe `nemo::io::http`): these are the `unpack_…` and `validate_…` functions, which are currently methods, even though there is no reason for them to be, as well as the `as_lexical_value` function, which currently validates the value domain for HTTP parameters (that part should go along the other validation logic) and also returns `value.lexical_value()` (that shouldn't be part of that function at all, rather a separate call).

mmarx:
The CI failures might be fixed by bumping `oxttl` (i.e., by running `cargo update`)

nemo-physical/src/resource.rs
```diff
@@ -94,383 +96,349 @@ impl Resource {
...
-    fragment: Option<String>,
-    post_parameters: Vec<(String, String)>,
+    resource: Resource,
+    header_names: HashSet<String>,
 }
 
 impl ResourceBuilder {
     /// Returns the path of a local iri
-    pub fn strip_local_iri(iri: Iri<String>) -> Result<String, ResourceValidationErrorKind> {
+    fn strip_local_iri(iri: Iri<String>) -> Result<String, ResourceValidationErrorKind> {
```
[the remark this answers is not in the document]
Laschoking: The method is used in `ResourceBuilder::try_from()`. So I can move it in the module level of the `resource` file, but moving it to`nemo::io` would cause dependency issues.

nemo-physical/src/resource.rs
```diff
@@ -94,383 +96,349 @@ impl Resource {
...
+                    })
+                })
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+        Ok(result.into_iter())
+    }
+
+    /// Try to convert a single HTTP parameter value into  a [String]
+    fn as_lexical_value(
+        &self,
```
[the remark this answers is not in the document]
Laschoking: moved function to module as `validate_http_parameter_value_domain(value: &AnyDataValue)`

nemo/src/io/http_parameters.rs
```diff
@@ -0,0 +1,112 @@
...
+    if parameters.value_domain() != ValueDomain::Map {
+        return Err(ValidationErrorKind::HttpParameterNotInValueDomain {
+            expected: (vec![parameters.value_domain()]),
+            given: (ValueDomain::Map),
+        });
+    }
+    let keys = parameters.map_keys().expect("Verified to be a map");
+    for key in keys {
+        validate_http_parameter_value_domain(key)?;
+    }
```
mmarx, line 77:
```rust
parameters.map_keys().expect("Verified to be a map").try_for_each(validate_http_parameter_value_domain)?;
```

nemo-physical/src/error.rs
```diff
@@ -40,6 +40,9 @@ pub enum ReadingErrorKind {
     /// Error in Reqwest's HTTP handler
     #[error(transparent)]
     HttpTransfer(#[from] reqwest::Error),
+    /// Error when HTTP response has not the expected content type
+    #[error("HTTP response failed")]
+    HttpWrongContentType,
```
mmarx, line 45: We no longer use this, do we?

mmarx (changes requested):
Just some nitpicks

nemo-physical/src/resource.rs
```diff
@@ -352,3 +346,78 @@ impl TryFrom<AnyDataValue> for ResourceBuilder {
...
+        let err = builder.add_header(key.clone(), value).unwrap_err();
+        assert_eq!(err, ResourceValidationErrorKind::DuplicateHttpHeader(key));
+
+        Ok(())
+    }
+
+    #[test]
+    fn create_path_resource() {
+        let builder = TryInto::<ResourceBuilder>::try_into(String::from(
+            "file:directory/file.extension#fragment",
```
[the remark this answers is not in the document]
Laschoking: I changed the test to `file:/directory`, but currently `@import test :- csv{resource=<file:test.csv>, format=(string,)}.
` is still valid and will be resolved as relative path with resource `Path(test.csv)`.

nemo/src/io/format_builder.rs
```diff
@@ -231,7 +231,9 @@ impl<Tag> FormatParameter<Tag> for StandardParameter {
 
         match self {
             StandardParameter::BaseParamType(no_parameters) => match *no_parameters {},
-            StandardParameter::Resource => Ok(()),
+            StandardParameter::Resource => ResourceBuilder::try_from(value)
+                .map(|_| ())
```
mmarx, line 235: we're using `.and(Ok(()))` elsewhere, would be nice to stay consistent (also several times below)

nemo-physical/src/resource.rs
```diff
@@ -339,14 +339,16 @@ impl TryFrom<AnyDataValue> for ResourceBuilder {
...
 
     fn try_from(value: AnyDataValue) -> Result<Self, Self::Error> {
-        value
-            .to_plain_string()
-            .or_else(|| value.to_iri())
-            .ok_or(ResourceValidationErrorKind::InvalidResourceFormat {
+        match value.value_domain() {
+            ValueDomain::PlainString => value.to_plain_string_unchecked().try_into(),
+            ValueDomain::Iri => Iri::parse(value.to_iri_unchecked())
+                .map_err(|_| ResourceValidationErrorKind::InvalidIri)?
```
mmarx, line 345: It'd probably be good if we don't throw away the error here, since it'll contain information on which part is invalid?

mmarx (approved):

PR #625, opened 2025-02-27
Add builtin for computing the Levenshtein distance of two strings

mmarx:
This adds the `LEVENSHTEIN(x, y)` builtin, which computes the [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) between the strings `x` and `y`.


monsterkrampe (approved):
LGTM :)

PR #626, opened 2025-03-03
Replace deprecated rio with oxrdfio

### LINKED ISSUE #623: Malformed Prefixes in Turtle Files Cause Crashing
### Description  
Providing certain malformed `@prefix` directives in the resource file causes nemo to crash. 

### Steps to Reproduce  
1. `mwe.rls` file
    ```prolog
    @import TRIPLE :- turtle { resource = "triples.ttl" } .
    @export TRIPLE :- turtle {}.
    ```
2. `triples.ttl` file:  
    ```turtle
    # @prefix : <#>.
    # @prefix ont: <>.
    # @prefix riskm: <./>.
    @prefix owl: <http://www.w3.org/2002/07/owl#>.

    [ a owl:Ontology ] .
    ```
3. Uncomment any of the commented `@prefix` directives.  
4. Run `nmo mwe.rls`.

mmarx:
Replaces our dependency on the (deprecated by upstream) `rio` crates, in favour of `oxrdfio`. This also fixes #623, as `oxrdfio` is more robust in the face of malformed input.

### DIFF
--- a/nemo/src/io/formats/rdf/reader.rs
+++ b/nemo/src/io/formats/rdf/reader.rs
@@ -170,7 +170,10 @@ impl RdfReader {
         let stop_limit = self.limit.unwrap_or(u64::MAX);
         let triple_count = Cell::new(0);
 
-        let mut on_triple = |triple: Quad| {
+        // [RdfParser] always returns quads, but we ignore the graph
+        // name here (since it will always be
+        // [GraphName::DefaultGraph]).
+        let mut on_quad = |quad: Quad| {
             // This is needed since a parser might process several RDF statements
             // before giving us a chance to stop in the outer loop.
             if triple_count.get() == stop_limit {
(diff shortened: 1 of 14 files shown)


nemo/src/io/formats/rdf/reader.rs
```diff
@@ -184,7 +170,7 @@ impl RdfReader {
         let stop_limit = self.limit.unwrap_or(u64::MAX);
         let triple_count = Cell::new(0);
 
-        let mut on_triple = |triple: Triple| {
+        let mut on_triple = |triple: Quad| {
```
monsterkrampe, line 173: Just double checking: `Triple` is `Quad` now?

mmarx: Yes, `RdfParser` always iterates over Quads, since it doesn't know up front which format it will parse.

monsterkrampe: Ok, then I think we should also rename variables accordingly to prevent confusion.

mmarx: I hope this is clearer now.

monsterkrampe (approved):

nemo/src/io/formats/rdf/reader.rs
monsterkrampe: Thanks!

PR #630, opened 2025-03-20
Include rule attributes in json trace

### LINKED ISSUE #627: Rule attributes in trace file
It would be super useful if rule attributes are also exported into the trace file.
I.e. the rule names of the corresponding inferences in the json trace file contain the defined display name.
Like so: 

rule in .rls file:

```
#[name("TRANSIT")]
r(x, z) :- r(x, y), r(y, z) .

```

inference in json trace file using said rule
```
ruleName: "TRANSIT: r(x, z) :- r(x, y), r(y, z) ."
conclusion: r(a,c)
premises:
     0: r(a,b)
     1: r(b,c)

```

aannleax:
This PR adds rule attributes (i.e. `#[name("rule name")]`, `#[display(f"result: {?x}")]`) into the json output format for tracing.

For example:
```
{
    "finalConclusion": [
        "ancestor(alice, gilbert)"
    ],
    "inferences": [
        {
            "conclusion": "ancestor(alice, gilbert)",
            "premises": [
                "ancestor(alice, edward)",
                "parent(edward, gilbert)"
            ],
            "rule": "ancestor(?X, ?Z) :- ancestor(?X, ?Y), parent(?Y, ?Z) .",
            "ruleDisplay": "alice, edward, gilbert -> alice, gilbert",
            "ruleName": "Ancestor Rule"
        },
     ...
    ]
}
```

Closes #627.

### DIFF
--- a/nemo-python/tests/test_example.py
+++ b/nemo-python/tests/test_example.py
@@ -29,6 +29,9 @@ class TestExample(unittest.TestCase):
         interesting(msg).
 
         interesting(?x) :- data(?x, ?y), interesting(?y).
+
+        #[name("interesting rule")]
+        #[name(f"data: {?x}, {?y}")]
         interesting(?y) :- data(?x, ?y), interesting(?x).
         """
 
@@ -89,6 +92,8 @@ class TestExample(unittest.TestCase):
         expected_trace = {
             "rule": "interesting(?y) :- data(?x, ?y), interesting(?x) .",
             "assignment": {"?x": 3.14, "?y": "<circle>"},
+            "name": "interesting rule",
+            "description": "data: 3.14, <circle>",
             "subtraces": [
                 {
                     "fact": 'data("3.14"^^<http://www.w3.org/2001/XMLSchema#double>, circle)'
@@ -96,6 +101,8 @@ class TestExample(unittest.TestCase):
                 {
                     "rule": "interesting(?y) :- data(?x, ?y), interesting(?x) .",
                     "assignment": {"?x": "<py>", "?y": 3.14},
+                    "name": "interesting rule",
+                    "description": "data: <py>, 3.14",
                     "subtraces": [
                         {
                             "fact": 'data(py, "3.14"^^<http://www.w3.org/2001/XMLSchema#double>)'
--- a/nemo-python/tests/test_example.py
+++ b/nemo-python/tests/test_example.py
@@ -31,7 +31,7 @@ class TestExample(unittest.TestCase):
         interesting(?x) :- data(?x, ?y), interesting(?y).
 
         #[name("interesting rule")]
-        #[name(f"data: {?x}, {?y}")]
+        #[display(f"data: {?x}, {?y}")]
         interesting(?y) :- data(?x, ?y), interesting(?x).
         """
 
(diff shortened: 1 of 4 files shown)


nemo/src/rule_model/components/rule.rs
```diff
@@ -78,30 +78,21 @@ impl Rule {
...
+        self.name.clone()
+    }
+
     /// Return a string representation of the rule instantiated with the given [Substitution].
-    /// This will either return
-    ///     * The content of the display attribute for this rule
-    ///     * a canonical string representation of the rule (i.e. [Display] representation)
-    /// whichever is the first defined in this list.
-    pub fn display_instantiated(&self, substitution: &Substitution) -> String {
+    pub fn display_instantiated(&self, substitution: &Substitution) -> Option<String> {
```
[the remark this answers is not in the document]
aannleax: The description did change. It now just returns the #[display(...)] string instantiated with the given substitution (so the name now fits better than before). The previous behavior was very specific to the ascii tree representation of the trace and has been therefore moved there (see `fn to_instantiated_string(&self) -> String` in trace.rs)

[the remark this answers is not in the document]
aannleax: Now changed "display" to "description"

nemo/src/execution/tracing/trace.rs
```diff
@@ -342,20 +364,32 @@ impl ExecutionTrace {
...
-    rule_name: String,
+    /// String representation of the logical rule
+    #[serde(rename = "rule")]
+    rule_string: String,
+    /// Optionally provided name of the rule
+    #[serde(rename = "ruleName", skip_serializing_if = "Option::is_none")]
+    rule_name: Option<String>,
+    /// Optional human readable string representation of the rule
+    #[serde(rename = "ruleDisplay", skip_serializing_if = "Option::is_none")]
+    rule_display: Option<String>,
```
monsterkrampe, line 389: I think this should be renamed to `rule_description`.

nemo/src/rule_model/components/rule.rs
```diff
@@ -74,34 +74,27 @@ impl Rule {
...
+    /// Return the name of the rule, if it is given one.
+    pub fn name(&self) -> Option<String> {
+        self.name.clone()
+    }
+
+    /// Return a string representation of the rule's description with the given [Substitution].
+    ///
+    /// Returns `None` if `description` results not in a ground term after applying the substitution.
+    pub fn description_instantiated(&self, substitution: &Substitution) -> Option<String> {
+        if let Some(mut display) = self.description.clone() {
```
monsterkrampe, line 90: The local variable should likely also be called `description` to avoid confusion.

nemo-python/tests/test_example.py
```diff
@@ -29,6 +29,9 @@ def setUp(self):
         interesting(msg).
 
         interesting(?x) :- data(?x, ?y), interesting(?y).
+
+        #[name("interesting rule")]
+        #[display(f"data: {?x}, {?y}")]
```
monsterkrampe, line 34: This should also be `description`, no?

monsterkrampe (changes requested):
Thanks for clarification and renaming. It seems though that there are a few places where `display` still needs to be replaced by `description`.

aannleax:
Maybe let's just keep it simple and move back to `display` ^^

The function name of the getter is now `instantiated_display`

monsterkrampe:
This is also fine by me :)

monsterkrampe (approved):

PR #649, opened 2025-03-27
Remove incorrect warning about the content-length of http responses

### LINKED ISSUE #648: Http content length warning behaves unexpected
The check below from `http.rs` does not work as expected:
```
if response
            .headers()
            .get(reqwest::header::CONTENT_LENGTH)
            .is_none()
        {
            warn!("HTTP response to the request: \n {full_url} \n contains no content. Make sure you specified the correct IRI and parameters.");
        }
```
The HTTP response for a valid request will only contain the `CONTENT_LENGTH` header, if the GET request was cached already. If the request is valid but not cached no content length is provided which results in an unexpected warning.

Therefore, the warning should be removed.

Laschoking:
Resolves #648 .

### DIFF
--- a/nemo/src/io/resource_providers/http.rs
+++ b/nemo/src/io/resource_providers/http.rs
@@ -128,6 +128,10 @@ impl HttpResourceProvider {
         // do any character set guessing, as `response.text()` would do.
         let content = response.bytes().await.map_err(err_mapping)?;
 
+        if content.is_empty() {
+            warn!("HTTP response to the request: \n {full_url} \n contains no content. Make sure you specified the correct IRI and parameters.");
+        }
+        
         Ok(HttpResource {
             resource: resource.clone(),
             content: content.into(),
--- a/nemo/src/io/resource_providers/http.rs
+++ b/nemo/src/io/resource_providers/http.rs
@@ -123,13 +123,6 @@ impl HttpResourceProvider {
         if !response_type.contains(media_type) {
             warn!("HTTP response to the request: \n {full_url} \n is of content type {response_type:?}, expected {media_type:?}. Make sure you specified the correct IRI and parameters.");
         }
-        if response
-            .headers()
-            .get(reqwest::header::CONTENT_LENGTH)
-            .is_none()
-        {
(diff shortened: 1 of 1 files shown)


mmarx:
Can't you just move this a bit further down and check `content.is_empty()` after https://github.com/knowsys/nemo/blob/8299dd0d198752f956addbe05e49bd5dd62d13a6/nemo/src/io/resource_providers/http.rs#L136

Laschoking:
in case of an endpoint side error  `content` is not empty but contains html: `b"<!DOCTYPE html><html lang=\"en\" dir=\"ltr\"><head><meta charset=\"utf-8\"><meta ... `

mmarx:
Sure, but that won't be reached if an error occurs anyways, since then the status code wouldn't be 200. It would also not have worked with the old approach (since the content-length for an error page is also non-zero), and I don't think that this is the case the warning should try to highlight, anyways: we want to warn if we get an empty response with a 200 status code, and the place to check for that is after line 136, where we have the full response available. We're already erroring on a non-200 status code, and we're warning if the content-type doesn't fit (which should trigger for an HTML error page, e.g.).

Laschoking:
I agree that we want to have an empty content response warning. I just don't see how this can be achieved. 
The response has a 200 status code even if the endpoint is not accurate e.g., `<https://query.wikidata.org>` instead of `<https://query.wikidata.org/sparql>` and in this case the response is not empty. Maybe the warning should rather happen in the reader?

mmarx:
That … is just not a case that this warning will be able to catch, irregardless of where we place it. This should warn if the status code is 200 and the response is empty, because that is surprising behaviour that won't be caught by the reader. It will not catch all cases of mistyping the endpoint, and that is okay. It can also be triggered by network connectivity problems, or by the endpoint itself failing, and that is also okay. Specifically, using `http://query.wikidata.org` over `https://query.wikidata.org/sparql` _will_ trigger the content-type warning, and that's good enough to alert the user, why would we want to have _another_ warning for the same case?

Laschoking:
I guess this was a little bit of a misunderstanding, I initially thought we were talking about Header validation. I will add the `content.is_empty()` check.

mmarx (approved):

PR #654, opened 2025-04-01
Fix GZip decoding

### LINKED ISSUE #652: GZip decoding stopped working
When running the following program:

```
@import data :- csv {resource = "data.csv.gz"}.

result(?x, ?y) :- data(?x, ?y).
```

given a gzipped non-empty input `data.csv.gz`, I get no result:

```
Reasoning completed in 1ms. Derived 0 facts.
   Data import:     1ms
   Reasoning:       0ms
   Data export:     0ms
```

mmarx:
Fixes #652.


mmarx:
Two different things got broken as part of #621:
 * detecting the compression from the resource name was only done for RDF formats, not all data sources
 * extracting the file extension from IRI resources indexed into the wrong slice (the whole IRI, rather than just the path portion that the index was relative to – although the index was also off-by-one)

monsterkrampe:
I see, thanks for the clarification :)

PR #655, opened 2025-04-07
Feature/import stdin

### LINKED ISSUE #622: Support reading one import from stdin
For some uses of Nemo, it would be nice to pass data through stdin. Currently, one would have to write data to a file to pass it to Nemo. 

Nemo already supports the use of `"-"` as a resource in `@export` to write to stdout, and the same syntax could work for `@import` and stdin. In addition, Nemo should check that at most one stdin source is used.

Laschoking:
This resolves #622 .
The error `ReachedStdinImportLimit` is caused if at least two import statements include `resource=""` e.g.,
```
@import foo :- csv{resource="",format=(int,int)}.
@import bar :- csv{resource="",format=(int,int)}.
```

### DIFF
--- a/nemo-physical/src/resource.rs
+++ b/nemo-physical/src/resource.rs
@@ -78,7 +78,7 @@ impl Resource {
     /// Whether the resource supports compression
     pub fn supports_compression(&self) -> bool {
         match self {
-            Resource::Pipe => false,       // never compress output to standard out
+            Resource::Pipe => false, // never compress input from standard in or output to standard out
             Resource::Http { .. } => true, // reqwest handles transport layer compression, but retrieved resources may still need decompression
             Resource::Path(_) => true,     // for local files, we need to handle compression
         }
@@ -175,7 +175,7 @@ impl fmt::Display for Resource {
                 }
             }
             Self::Path(path) => write!(f, "{}", path.display()),
-            Self::Pipe => f.write_str("stdin/stdout"),
+            Self::Pipe => f.write_str("standard in / standard out"),
         }
     }
 }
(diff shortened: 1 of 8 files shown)


nemo-physical/src/resource.rs
```diff
@@ -94,9 +94,9 @@ impl Resource {
         matches!(self, Self::Path(..))
     }
 
-    /// Return if resource points to stdout
-    pub fn is_stdout(&self) -> bool {
-        matches!(self, Self::Stdout)
+    /// Return if resource points to pipi
```
mmarx, line 97: `pipe`

nemo/src/rule_model/error/validation_error.rs
```diff
@@ -184,10 +184,10 @@ pub enum ValidationErrorKind {
         /// The actual [ValueDomain] of the HTTP parameter
         given: ValueDomain,
     },
-    /// Stdout is not supported for imports
-    #[error("resource `stdout` is not supported for imports")]
+    /// Stdin is only supported for one import
+    #[error("Expected at most one `stdin` import, found at least 2 occurrences")]
```
mmarx, line 188: If I recall correctly, error messages are supposed to start with lowercase?

nemo/src/rule_model/program.rs
```diff
@@ -239,9 +239,19 @@ impl Program {
...
         let mut predicate_arity = HashMap::<Tag, (usize, Origin)>::new();
 
+        let mut count_stdin = 0;
         for (import_directive, import_builder) in self.imports() {
             let predicate = import_directive.predicate().clone();
             let origin = *import_directive.origin();
+            if matches!(
+                import_builder.resource().map(|resource| resource.is_pipe()),
+                Some(true)
+            ) {
```
mmarx, line 249:
```rust
if import_builder.resource().is_some_and(|resource| resource.is_pipe()) {
```

mmarx (changes requested):
Some minor changes, but looks good otherwise

nemo/src/rule_model/error/validation_error.rs
Laschoking: Yes, you're right!

mmarx (approved):

PR #656, opened 2025-04-15
Round negative floats and doubles towards 0.0

### LINKED ISSUE #634: ROUND up for negative values ending in .5
If SPARQL compliance of functions is a goal, the ROUND function should round up for negative values ending in .5.

```
input(-2.5) .  
output(ROUND(?i)) :- input(?i) .
```
Expected: -2.0
Got: -3.0

This exact example is from the SPARQL standard: https://www.w3.org/TR/sparql11-query/#func-round

Laschoking:
This resolves #634 by adding a special case that checks if the fractional part of a float or double equals `-0.5`. If this occurs `ceil()` is used instead of `round()`.

### DIFF
--- a/nemo-physical/src/datatypes/double.rs
+++ b/nemo-physical/src/datatypes/double.rs
@@ -79,9 +79,13 @@ impl Double {
     /// Returns the nearest integer to `self`.
     /// If a value is half-way between two integers, round away from 0.0.
     pub(crate) fn round(self) -> Self {
-        Double::new(self.0.round()).expect("operation returns valid float")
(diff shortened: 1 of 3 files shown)


nemo-physical/src/datatypes/double.rs
```diff
@@ -79,9 +79,13 @@ impl Double {
     /// Returns the nearest integer to `self`.
     /// If a value is half-way between two integers, round away from 0.0.
```
aannleax, line 80: The description should change accordingly

aannleax (changes requested):

aannleax (approved):

PR #659, opened 2025-04-16
Display error if ground operation is invalid

### LINKED ISSUE #658: Bug: Invalid string functions cause panic when used in facts
Several invalid string functions cause a panic:
```
thread 'main' panicked at nemo/src/chase_model/translation/fact.rs:34:17:
internal error: entered unreachable code: invalid program: fact contains non-primitive values
```

Example cases:
```
c(URIDECODE(Hello%2GWorld)). % Wrong encoding
c(SUBSTR("12345", 7)). % start > string length
c(STRLEN(123)). % argument not a string

```
The functions return `None` which is not handled correctly.
If a body atom exists e.g., `c(SUBSTR("12345", 7)) :- a(_).` no panic is caused.

Solution: -> An Error message should be displayed if a user mistyped a fact.

Laschoking:
Verify the result of ground operations during validation process to display a `ValidationErrorKind::InvalidGroundOperation` error message.
Resolves #658.

### DIFF
--- a/nemo/src/rule_model/components/term/operation.rs
+++ b/nemo/src/rule_model/components/term/operation.rs
@@ -271,6 +271,11 @@ impl ProgramComponent for Operation {
             return None;
         }
 
+        if self.is_ground() && matches!(self.reduce(), Term::Operation(_)) {
+            builder.report_error(self.origin, ValidationErrorKind::InvalidGroundOperation);
+            return None;
+        }
+
         for argument in self.arguments() {
             argument.validate(builder)?;
         }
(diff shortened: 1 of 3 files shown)


nemo/src/rule_model/components/term/operation.rs
```diff
@@ -271,6 +271,11 @@ impl ProgramComponent for Operation {
             return None;
         }
 
+        if self.is_ground() && matches!(self.reduce(), Term::Operation(_)) {
```
aannleax, line 274: Maybe `self.reduce().is_primitive()` is a bit more robust?

aannleax (changes requested):

nemo/src/rule_model/components/term/operation.rs
Laschoking: good point. i replaced `matches!()` with `!self.reduce().is_primitive()`

nemo/src/rule_model/error/validation_error.rs
```diff
@@ -188,6 +188,10 @@ pub enum ValidationErrorKind {
     #[error("expected at most one `stdin` import, found at least 2 occurrences")]
     #[assoc(code = 237)]
     ReachedStdinImportLimit,
+    /// Ground operation contains invalid literals
+    #[error("ground operation contains at least one invalid literal")]
```
aannleax, line 192: looking at the error message, maybe it should mention that the operation does not return a result (instead of being "invalid")

aannleax (approved):

PR #664, opened 2025-04-23
add external filters

### LINKED ISSUE #568: Add Hook for Filtering Atoms
For a student project, we need a feature to be able to filter derived facts according to some (quite involved) function.
It would be nice, if we could just pass this to the execution engine to be applied as another filter on top of the derived facts.
Basically, if `l` is the list of derived facts, we want to call `l.filter(p)` where `p : Fact -> Bool` is our predicate function that we pass to the execution engine.
That should already be all we need. I hope the description is clear enough :)

@aannleax It would help me, if you can create some stubs for functions that would be required for this. I'm not sure where the best place to inject this would be. I can then fill in the gaps if you want me to :)
Also feel free to add the appropriate labels to this issue since I'm not sure which layers would actually be affected here.

matzemathics:
This might be useful for mfa checks

### DIFF
--- a/nemo/src/chase_model/components/rule.rs
+++ b/nemo/src/chase_model/components/rule.rs
@@ -1,5 +1,7 @@
 //! This module defines [ChaseRule].
 
+use nemo_physical::util::hook::FilterHook;
+
 use crate::rule_model::{
     components::{
         term::primitive::{variable::Variable, Primitive},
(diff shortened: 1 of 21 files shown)


nemo/src/chase_model/components/rule.rs
```diff
@@ -138,6 +142,11 @@ impl ChaseRule {
     pub(crate) fn aggregate_head_index(&self) -> Option<usize> {
         self.head.aggregate_head_index
     }
+
+    /// Return the filter hook.
+    pub(crate) fn filter_hook(&self) -> Option<FilterHook> {
```
matzemathics, line 147: this should return a reference

PR #667, opened 2025-04-24
Feature/636 lang tagged string functions

### LINKED ISSUE #636: Support language tagged strings in functions
Many functions do not support language tagged strings.

```
input("a") .  
input("bb"@en) .  
output(STRLEN(?i)) :- input(?i) .
```
Expected: 1, 2
Got: 1

Note that some functions have rules what combinations of language tags are allowed, but for functions like `STRLEN` it should be straight forward.

Laschoking:
String functions now support language tagged strings and follow  the SPARQL recommendations.
This does not change the functionality of the string functions, but rather defines in which cases language tags should be applied to the result or in which cases the operation should be rejected.

All Binary functions that expect two strings follow the Argument Compatibility Rules corresponding to https://www.w3.org/TR/sparql11-query/#func-arg-compatibility. Among other things, this ensures that two parameters with different, non-empty language tags are rejected.

`StringConcatenation` is not affected from this. Here, the result receives a  language tag if all parameters have an identical tag, otherwise no language tag is used.

Furthermore, the behavior of `StringSubstring` and `StringSubstringLength` is changed for edge cases to be conform with the other string functions. The functions will now return an empty string instead of `None`, for example:
`SUBSTR("abc", 5)`   -> results in  `""` instead of None
`SUBSTR("abc"@en, 5)`   -> results in  `""@en` instead of None
`SUBSTR("abc", 1, -1)`    -> results in  `""` instead of None
`SUBSTR("abc"@en, 1, -1)`  -> results in  `""@en` instead of None

This resolves #636.


nemo-physical/src/function/definitions/string.rs
```diff
@@ -90,15 +140,27 @@ impl BinaryFunction for StringCompare {
...
     fn evaluate(&self, parameters: &[AnyDataValue]) -> Option<AnyDataValue> {
-        string_vec_from_any(parameters)
-            .map(|strings| AnyDataValue::new_plain_string(strings.concat()))
+        let lang_strings = string_vec_from_any(parameters)?;
+        let result = lang_strings
+            .iter()
+            .map(|ls| ls.string.as_str())
+            .collect::<String>();
+
+        let mut iter = lang_strings.into_iter().map(|ls| ls.tag);
```
aannleax, line 158: Probably a good opportunity for `flat_map`

nemo-physical/src/function/definitions/string.rs
```diff
@@ -115,6 +177,8 @@ impl NaryFunction for StringConcatenation {
 /// Returns `true` from the boolean value space if the string provided as the second parameter
 /// is contained in the string provided as the first parameter and `false` otherwise.
 ///
+/// This function ignores language tags.
+///
```
aannleax, line 181: This is not correct anymore, right?

Meaning that incompatible language tags will result in an "error".

nemo-physical/src/function/definitions/string.rs
```diff
@@ -184,6 +258,8 @@ impl BinaryFunction for StringStarts {
 /// Returns `true` from the boolean value space if the string provided as the first parameter
 /// ends with the string provided as the second parameter and `false` otherwise.
 ///
+/// This function ignores language tags.
```
aannleax, line 261: See above

nemo-physical/src/function/definitions/string.rs
```diff
@@ -223,6 +305,10 @@ impl BinaryFunction for StringEnds {
 /// Returns the part of the string given in the first parameter which comes before
 /// the string provided as the second parameter.
 ///
+/// Applies the language tag of the first parameter on the result, if:
+/// - the resulting string is not empty
```
aannleax, line 309: I don't see such check nor do I see that condition described.

nemo-physical/src/function/definitions/string.rs
```diff
@@ -232,15 +318,28 @@ impl BinaryFunction for StringBefore {
...
+                                .graphemes(true)
+                                .collect::<Vec<&str>>()[..i]
+                                .join("");
+                            LangTaggedString::new(string, first_lang_string.tag)
+                        }
+                        None => {
+                            let tag = if second_lang_string.string.is_empty() {
+                                first_lang_string.tag
+                            } else {
+                                None
```
aannleax, line 337: I don't think this is correct, shouldn't it just be the empty string with the language tag of `first`?

Does `unicode_find(&first_lang_string.string, &second_lang_string.string)` return `None` when the second parameter is empty?

nemo-physical/src/function/definitions/string.rs
```diff
@@ -259,6 +358,10 @@ impl BinaryFunction for StringBefore {
 /// Returns the part of the string given in the first parameter which comes after
 /// the string provided as the second parameter.
 ///
+/// Applies the language tag of the first parameter on the result, if:
+/// - the resulting string is not empty
+/// - or the second parameter is empty
+///
 /// Returns `None` if either parameter is not a string.
 #[derive(Debug, Copy, Clone)]
 pub struct StringAfter;
```
aannleax, line 367: Similar as for `StringBefore`

nemo-physical/src/function/definitions/string.rs
```diff
@@ -341,6 +459,8 @@ static REGEX_CACHE: OnceCell<Mutex<lru::LruCache<String, regex::Regex>>> = OnceC
 ///
 /// Returns `None` if either parameter is not a string or the second parameter is not
 /// a regular expression.
+///
+/// This function ignores language tags.
```
aannleax, line 463: Again, except for compatibility

nemo-physical/src/function/definitions/string.rs
```diff
@@ -490,14 +632,19 @@ impl UnaryFunction for StringLowercase {
...
 impl UnaryFunction for StringUriEncode {
     fn evaluate(&self, parameter: AnyDataValue) -> Option<AnyDataValue> {
-        parameter
-            .to_plain_string()
-            .map(|string| AnyDataValue::new_plain_string(urlencoding::encode(&string).to_string()))
+        LangTaggedString::try_from(parameter)
+            .ok()
+            .map(|lang_string| {
+                let uri_encode = urlencoding::encode(&lang_string.string).to_string();
+                LangTaggedString::new(uri_encode, None).into_data_value()
```
aannleax, line 646: You can just construct an `AnyDataValue` directly here

nemo-physical/src/function/definitions/string.rs
```diff
@@ -513,15 +660,16 @@ impl UnaryFunction for StringUriEncode {
...
 pub struct StringUriDecode;
 impl UnaryFunction for StringUriDecode {
     fn evaluate(&self, parameter: AnyDataValue) -> Option<AnyDataValue> {
-        let string = parameter.to_plain_string()?;
-        let decoded = urlencoding::decode(&string).ok()?;
-
-        Some(AnyDataValue::new_plain_string(decoded.to_string()))
+        let lang_string = LangTaggedString::try_from(parameter).ok()?;
+        let uri_decode = urlencoding::decode(&lang_string.string).ok()?.to_string();
+        Some(LangTaggedString::new(uri_decode, None).into_data_value())
```
aannleax, line 672: You can just construct an `AnyDataValue` directly here

nemo-physical/src/function/definitions/string.rs
Laschoking: The current implementation takes all LangTags (containing possibly `None`) and then checks if all LangTags in the iterator match with the first LangTag. If the first LangTag was `None` so will be the LangTag of the result.
 Using `flat_map` would require a small change:
```
        let mut iter = lang_strings.into_iter().flat_map(|ls| ls.tag);
        let lang_params = iter.clone().count();
        let result_tag = if lang_params == parameters.len() {
        // Each Param has  a lang tag
            iter
            .next()
            .filter(|first| iter.all(|tag| &tag == first))
        }
        else {
            None
        };
```
But maybe this is more readable than the first version

nemo-physical/src/function/definitions/string.rs
Laschoking: Maybe it is incorrectly formulated. Basically, all binary functions that require 2 strings will only return a result, if both strings are compatible, and otherwise result in an error.
Once this is checked some string function like  `STRBEFORE, STRAFTER` use the compatible lang tags for their result, while others like `STRINGCONTAINS` will just return a boolean, so no language tag is applied.
I can also remove this kind of comments that just state that they don't use validated lang tags any further.

Laschoking: I reformulated most comments :)

aannleax (approved):
Look good now :)

PR #669, opened 2025-04-25
Fix Bug/620 duplicate aggregate variables

### LINKED ISSUE #620: Aggregate rule panics, if aggregate variable and distinct variable are the same
The following rule
```prolog
r(#MAX(?x, ?x)) :- p(?x).
```

leads to a panic:
```
thread 'main' panicked at nemo-physical/src/util/mapping/ordered_choice.rs:37:9:
assertion failed: source.len() >= target.len()
```

This should probably be rejected by program analysis.

Laschoking:
Ensures that aggregates only consider the distinct variables for computation.
If a primitive variable appears several times in an aggregate, a warning is displayed instead of an error.
Once warnings can be issued via the `error_builder`,  `log::warn!()` should be replaced with `error_builder.warning()` .
This resolves #620.


aannleax (approved):
Looks good. 

Should rebase on https://github.com/knowsys/nemo/pull/671 once its merged into main.

PR #671, opened 2025-04-25
Update some deps

mkroetzsch:
This branch is to update some outdated dependencies where it is painless to do so. Needs some more work to make Clippy happy.


mkroetzsch:
The one remaining issue is related to lifetime management in your code, Alex. I am not sure how to best fix this.

aannleax:
> The one remaining issue is related to lifetime management in your code, Alex. I am not sure how to best fix this.

Yeah this is because the `delegate` macro expands the code as follows:

```Rust
#[inline]
fn generate<'a>(
    &'_ self,
    input: Vec<Option<TrieScanEnum<'a>>>,
    dictionary: &'a RefCell<Dict>,
) -> Option<TrieScanEnum<'a>> {
    match self {
        Self::Aggregate(generator) => generator.generate::<'a>(input, dictionary),
        Self::Join(generator) => generator.generate::<'a>(input, dictionary),
        Self::Union(generator) => generator.generate::<'a>(input, dictionary),
        Self::Subtract(generator) => generator.generate::<'a>(input, dictionary),
        Self::Filter(generator) => generator.generate::<'a>(input, dictionary),
        Self::Function(generator) => generator.generate::<'a>(input, dictionary),
        Self::Null(generator) => generator.generate::<'a>(input, dictionary),
    }
}
```

Now every call has the redundant `::<'a>` which Rust doesn't like anymore. 

Not sure there is a way to fix this easily, so I would just ignore this warning for now and wait till it's fixed by the authors of the crate or alternatively just manually write that function.

aannleax (approved):
Apart from the one warning, looks good

Issue #670, opened 2025-04-25
CLI argument to dump facts


Laschoking:
The flag `--dump` could expect one of the  arguments `all`, `idb` or `edb`  and will then print the corresponding facts in stdout.
This option could simplify debugging or the generation of facts for tracing as in #668 .

Laschoking:
Resolved by #667

Issue #668, opened 2025-04-25
Getting traces for all inferred facts.


henrietteharmse:
Is there a way to get traces for all inferred facts?

Laschoking:
This feature does not exist yet. For now, you could use the `--trace-input-file` argument to read a list of facts.
So a work around would be:
1. export facts
```
data() :- ... % amazing nemo program
@export data :- csv{resource=" trace-data.csv"}.
```
2. Convert them to a tracable format:
Lets say ` trace-data.csv` looks like this:
```
"""a""",""b""",1
"""c""",""d""",2
```
you want this final format:
`data("a","b",1);data("c","d",2)`
3. Run the nemo again with `--trace-input-file trace-data.csv`

henrietteharmse:
@Laschoking Thanks. That is helpful to know!
BTW: love the "amazing nemo" comment :-)

PR #675, opened 2025-04-30
Feature/670 dump facts

Laschoking:
This resolves feature #670  and can be used to trace facts more easily as discussed in  #668.

### DIFF
--- a/nemo-cli/src/cli.rs
+++ b/nemo-cli/src/cli.rs
@@ -16,12 +16,33 @@ pub(crate) enum Exporting {
     Keep,
     /// Disable all exports.
     None,
-    /// Export all IDB predicates (those used in rule heads)
+    /// Export all IDB predicates (those used in rule heads).
     Idb,
-    /// Export all EDB predicates (those for which facts are given or imported)
+    /// Export all EDB predicates (those for which facts are given or imported).
     Edb,
     /// Export all predicates.
     All,
+}
+
+/// Possible settings for the fact-printing option.
+#[derive(clap::ValueEnum, Clone, Copy, Default, Debug, PartialEq, Eq)]
+pub(crate) enum FactPrinting {
+    /// Do not print facts for any predicate.
+    #[default]
+    None,
+    /// Print facts for all IDB predicates (those used in rule heads).
+    Idb,
+    /// Print facts for all EDB predicates (those for which facts are given or imported).
+    Edb,
+    /// Print facts for all predicates.
+    All,
+}
+
+impl FactPrinting {
+    /// Whether printing is enabled for some predicates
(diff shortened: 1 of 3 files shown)


nemo-cli/src/cli.rs
```diff
@@ -103,6 +103,12 @@ pub(crate) struct OutputArgs {
     /// does not affect export directives that already specify a compression
     #[arg(short, long = "gzip", default_value = "false")]
     gz: bool,
+    /// Specify which predicates to dump
+    #[arg(long = "dump", value_enum)]
+    pub(crate) dumping: Option<Dumping>,
```
monsterkrampe, line 108: I think I'd rather give this a name that clarifies that this is dumping facts.
For example, `dump-facts` might do it. Maybe even `print-facts` since this leads to writing the facts to stdout.
The internal naming should be changed accordingly but is not as critical as the user-facing argument names.

nemo-cli/src/cli.rs
```diff
@@ -103,6 +103,12 @@ pub(crate) struct OutputArgs {
...
+    /// File to dump the specified facts
+    #[arg(long = "dump-file", requires = "dumping")]
+    pub(crate) dump_file: Option<PathBuf>,
```
monsterkrampe, line 111: I don't think that this is needed as a user could simply redirect stdout to a file.

nemo-cli/src/main.rs
```diff
@@ -77,6 +93,28 @@ fn override_exports(program: &mut Program, value: Exporting) {
...
+                predicates.push(predicate);
+            }
+        }
+        Dumping::All => {
+            for predicate in program.all_predicates() {
+                predicates.push(predicate);
+            }
+        }
+    }
+    predicates
```
monsterkrampe, line 115: Why is this method not simply implemented along the lines of:
```rust
match dumping {
    Dumping::Idb => program.derived_predicates().collect(),
    ...
}

File: nemo-cli/src/main.rs [hunk 4]
```diff
@@ -298,7 +336,7 @@ fn run(mut cli: CliApp) -> Result<(), CliError> {
 
     let mut stdout_used = false;
 
-    if !export_manager.write_disabled() {
+    if !export_manager.write_disabled() || cli.output.dumping.is_some() {
```
monsterkrampe, line 339: Not dumping to stdout also activates the usual export even if this `export_manager.write_disabled()` is true. This does not seem right to me and is likely not expected by a user.
Could this not be completely separate?

monsterkrampe (changes requested):

monsterkrampe (approved):
LGTM, thanks for taking care of this @mmarx!

PR #682, opened 2025-05-21
Feature/transformations

aannleax:
Rework of the rule model to support program transformations


nemo-cli/src/cli.rs
```diff
@@ -204,8 +204,10 @@ pub struct CliApp {
 #[derive(Debug, Clone)]
 pub(crate) struct ParamKeyValue {
     /// Key: Global variable
+    #[allow(dead_code)] // TODO: Value is clearly read ?
```
mmarx, line 207: The warning occurs because the `xtask` crate depends on `nemo-cli` as a library (it needs the cli definition to generate manpages and completion functions), and there the value is indeed never read. The easy fix would be to just make the struct and its members `pub` instead of `pub(crate)`. The cleaner (but slighty more involved) fix would be to move `cli.rs` into its own library crate that is used by both `xtask` and `nemo-cli` (where it then also will need to be `pub`).

PR #686, opened 2025-05-26
Fix import of compressed RDF files via HTTP

mmarx:
Fixes HTTP imports of compressed RDF files, as in the following program, by correctly stripping the `.gz` before checking for a known RDF format:

```
@import TRIPLE :- rdf{resource="https://raw.githubusercontent.com/knowsys/nemo-examples/main/examples/owl-el/from-owl-rdf/galen-el.nt.gz"} .
```


monsterkrampe (approved):

monsterkrampe:
Thank you for fixing this!

PR #691, opened 2025-06-04
Build docker image with nix and push to ghcr

### LINKED ISSUE #689: Provide Docker Image for Nemo CLI
We are using Nemo in our pipelines. However, our users also want to run their own instances of our software. To make this easy for them, we provide a Docker image of our software.  Therefore, it will be really helpful if you can provide a Docker image for Nemo.

monsterkrampe:
Fixes #689 

This PR allows to build a docker image with `nix build .#docker`. On top, every commit on the main branch triggers the build of this image and pushes it to the Github Container Registry of this project with the `latest` tag.
Every git tag (aka. release) will also trigger a build tagging the docker image with the respective version (e.g. v.0.8.0 instead of latest). **This is currently not tested and I propose we check this when we do the next release.**

For example, the docker image from the registry can be used as follows:
```podman run --rm -t -v .:/pwd ghcr.io/knowsys/nemo:latest /pwd/test.rls -e idb -D /pwd/docker-results```
(The same should work with `docker` instead of `podman`.)
This already works right now as I did a test build/push from this branch tagging the image as `latest`. In the future this will only happen on `main`.

### DIFF
--- a/.github/workflows/nix-and-docker.yml
+++ b/.github/workflows/nix-and-docker.yml
@@ -1,4 +1,4 @@
-name: Build with nix
+name: Build with nix and create docker image
 
 on:
   push:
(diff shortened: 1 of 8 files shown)


monsterkrampe:
Updating the flake and fixing clippy issues now...

mmarx (approved):
LGTM; does indeed also work with podman (not that that should be any surprise, but it's what I had at hand).

PR #705, opened 2025-09-19
Proper async handling of http requests

monsterkrampe:


aannleax (approved):
No complaints from me :)

PR #709, opened 2025-09-24
Paging for SPARQL queries

mmarx:
Instead of pushing all of the bindings at once, split into multiple smaller queries if needed. This works with a static upper limit for the query size (currently 740 kb, which seems to _just_ fit for WDQS), and also properly handles `HTTP 413 Request Too Large` responses from the endpoint.

With this, the following program now successfully runs and fetches the given names for some 50000 humans in a total of 5 queries (one for the persons, and then 4 different sets of bindings).

```prolog
@prefix wdqs: <https://query.wikidata.org/> .
@prefix wd: <http://www.wikidata.org/entity/> .
@prefix wdt: <http://www.wikidata.org/prop/direct/> .

@import persons :- sparql{
  endpoint = wdqs:sparql,
  query = """
    PREFIX wd: <http://www.wikidata.org/entity/>
    PREFIX wdt: <http://www.wikidata.org/prop/direct/>
    SELECT ?s WHERE {
      ?s wdt:P31 wd:Q5
    } LIMIT 50000
  """
} .

@import names :- sparql{
  endpoint = wdqs:sparql,
  query = """
    PREFIX wdt: <http://www.wikidata.org/prop/direct/>
    SELECT ?s ?o WHERE {
      ?s wdt:P735 ?o
    }
  """
} .

% we need to proxy the persons, since import chains are currently not internalised.
people(?s) :- persons(?s) .

peopleNames(?s, ?n) :- people(?s), names(?s, ?n) .

@export peopleNames :- csv{} .
```

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -657,11 +657,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1752613317,
-        "narHash": "sha256-SPS/iE8RdqmlWeREDa9/k2dgziiaAJU3kDknuHYdGZk=",
+        "lastModified": 1758632847,
+        "narHash": "sha256-/BJTeCsBMx20S+N+V4aOd1O/LlKhr8BzpXuY6lfoTfs=",
         "owner": "knowsys",
         "repo": "nemo-vscode-extension",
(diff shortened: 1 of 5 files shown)


monsterkrampe (approved):
LGTM :)

PR #726, opened 2025-10-17
Feature/refactor planning

aannleax:
This PR replaces `ChaseRule` and `RuleAnalysis` with `NormalizedRule`s. It also reworks the planning code.

### DIFF
--- a/nemo/src/execution/planning/analysis/variable_order.rs
+++ b/nemo/src/execution/planning/analysis/variable_order.rs
@@ -988,7 +988,6 @@ mod test {
         ImportInstruction::new(predicate, handler)
     }
 
-    // #[ignore]
     #[test]
     #[cfg_attr(miri, ignore)]
     fn build_preferable_variable_orders_with_galen_el_part_ie_5_rules_without_constant() {
@@ -1000,6 +999,7 @@ mod test {
             let (predicate, arity) = predicates[predicate_index].clone();
             program.add_import(csv_import(predicate, arity));
         }
+
         for rule in rules {
             program.add_rule(rule);
         }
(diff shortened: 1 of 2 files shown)


nemo-physical/src/tabular/buffer/tuple_buffer.rs
```diff
@@ -477,18 +477,18 @@ impl TupleBuffer {
...
             for transformation in &pattern.transformations {
-                let value = transformation
-                    .program
-                    .evaluate_data(&data_values)
-                    .expect("should evaluate to a value");
-                self.current_tuple_data_values[transformation.position] = value;
+                if let Some(value) = transformation.program.evaluate_data(&data_values) {
+                    self.current_tuple_data_values[transformation.position] = value;
+                } else {
+                    continue 'pattern;
```
mmarx: I had to take a moment to think about why this would not evaluate to a value (when the data types do not match, e.g.) – a comment would be a good thing here

nemo/src/execution/planning/analysis/variable_order.rs
```diff
@@ -1262,12 +1277,13 @@ mod test {
         (rules, variables, predicates)
     }
 
+    #[ignore]
```
mmarx: Same question, why is this ignored?

nemo/src/execution/planning/analysis/variable_order.rs
```diff
@@ -960,24 +974,25 @@ mod test {
...
             Default::default(),
             arity,
             Arc::new(MockHandler),
         );
 
-        ChaseImport::new(predicate, handler)
+        ImportInstruction::new(predicate, handler)
     }
 
+    #[ignore]
```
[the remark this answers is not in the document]
aannleax: I wanted to ask @monsterkrampe if he could take a look at it later, as I don't understand how my changes would affect the outcome of the variable order computation (and if this would even be a problem)

PR #727, opened 2025-10-17
Deal with DSV files missing columns in certain rows

mmarx:
Instead of treating a missing column in one row in a DSV file as a hard error (which would lead to nothing being imported at all if it happened in the first row), ignore those in the `csv` crate and deal with it on a per-row basis.


aannleax (approved):
Looks good :)

Issue #747, opened 2025-11-20
Wrong proof tree returned on NEV query


mkroetzsch:
The [NEV issue #8](https://github.com/imldresden/nev/issues/8) appears to be a Nemo problem. Here is the original description on how to reproduce it in NEV.

Steps to reproduce:
[...]

PR #750, opened 2025-11-21
fix tracing computation of valid nodes

aannleax:
Fixes a bug in type 2 tracing where operations would not get applied correctly during the bottom-up phase.

Also addresses an issue where ids would be assigned inconsistently between the two tracing types (see #747).


monsterkrampe (approved):
Thanks for also extending the `expect` messages. Much appreciated!

PR #787, opened 2026-05-12
Added Support for default prefixes in SPARQL queries (#647)

### LINKED ISSUE #647: Feature: Support default prefixes in SPARQL queries
Currently each SPARQL query is required to contain all prefixes that are used within the query. While, for example, wikidata accepts a [list](https://www.wikidata.org/wiki/EntitySchema:E49) of common prefixes, oxigraph will fail during the validation process if prefixes are missing.

Writing several queries in one program that possibly share a big number of prefixes can therefore feel bulky and inconvenient for users.

According to https://github.com/oxigraph/oxigraph/issues/1177, oxigraph is planning to accept a collection of default prefixes when calling `Query::Parse()`.

Yesterday, we discussed two non-exclusive options:

1. Introduce a new import  parameter for sparql e.g., `default_prefixes={xsd, wikidata, ... }` . This would result in pushing a list of hard coded prefixes into the query before the oxigraph validation. 
 %% Note: In the validation process oxigraph will make all IRIs absolute and drop all prefixes.  If the same prefix name is supplied several times in the query, that latest one will be used during validation and expansion of IRIs.

2. Push the outer prefixes from the Nemo program into the query, for example
```
@prefix wd: <http://www.wikidata.org/entity/>.
@import test :- sparql{endpoint=<https://query.wikidata.org/sparql>, query= """SELECT * WHERE {?a ?b wd:Q123}""" }.
```

In my opinion option 1 is low effort but seems a little hacky. Once Oxigraph allows a list of default prefixes, we can also drop the manual pushing of hard coded prefixes into the query and just pass them to the parser directly.

Option 2 is nice to have and would especially come in handy when using custom prefixes.
Since I don't have much knowledge of the prefix expansion in the parser I am not sure how much effort this will take.

Aeternant:
Makes base and prefix IRIs available to Import/Export and closes #647


nemo/src/io/formats/dsv.rs
```diff
@@ -178,12 +180,17 @@ impl FormatParameter<DsvTag> for DsvParameter {
         }
     }
 
-    fn is_value_valid(&self, value: AnyDataValue) -> Result<(), ValidationError> {
+    fn is_value_valid(
+        &self,
+        value: AnyDataValue,
+        base: Option<Iri<String>>,
+        prefixes: HashMap<String, IriRef<String>>,
```
mmarx: This is fine for now, but we would want to consider moving these into a `FormatContext` (or something like that) struct if we ever add more parameters

nemo/src/io/formats/sparql.rs
```diff
@@ -93,13 +98,36 @@ impl FormatParameter<SparqlTag> for SparqlParameter {
...
+
+                let mut parser = SparqlParser::new();
+
+                let resolved_prefixes: Vec<(String, Iri<String>)> = prefixes
+                    .iter()
+                    .map(|(key, value)| {
+                        (
+                            key.clone(),
+                            Iri::from_str(&value.to_string())
+                                .unwrap_or_else(|_| base.clone().unwrap().resolve(value).unwrap()),
```
mmarx: We generally avoid `unwrap()` in favour of `expect()` (this is somewhat less strict in tests, but in library code, we don't want to have any plain `unwrap`s).

Here, I don't even think the `unwrap` on `base` is warranted – it might be `None` and that could still be valid if the relative prefix isn't actually used in the query. The validation should proceed like this:
 - if `base` is `None`, only collect the absolute prefixes (if the prefix is required in the query, the existing error handling will already catch it, if not, this is fine)
 - otherwise, collect all the prefixes, resolving relative prefixes with respect to the base (this will always succeed, so you can `base.resolve(value).expect("relative IRI can always be resolved with the absolute base")` here

```rust
let resolved_prefixes = if let Some(base) = base {
  prefixes.iter().map(|(key, value)| (key.clone(), Iri::from_str(value).unwrap_or_else(|_| base.resolve(value).expect("relative IRI can always be resolved with absolute base")).collect()
} else {
  prefixes.iter().filter_map(|(key, value)| (key.clone(), Iri::from_str(value)?).collect()
}
```

nemo/src/io/formats/sparql.rs
```diff
@@ -93,13 +98,36 @@ impl FormatParameter<SparqlTag> for SparqlParameter {
...
+                        )
+                    })
+                    .collect();
+
+                if let Some(base) = base {
+                    parser = parser.with_base_iri(base.to_string()).unwrap();
```
mmarx:
```suggestion
                    parser = parser.with_base_iri(base.to_string()).expect("IRI is valid");
```

nemo/src/io/formats/sparql.rs
```diff
@@ -93,13 +98,36 @@ impl FormatParameter<SparqlTag> for SparqlParameter {
...
+                }
+
+                for (prefix_name, prefix_iri) in resolved_prefixes {
+                    parser = parser
+                        .with_prefix(prefix_name, prefix_iri.to_string())
+                        .unwrap();
```
mmarx: (see above, also applies elsewhere in the PR)

nemo/src/io/formats/sparql.rs
```diff
@@ -164,7 +194,31 @@ impl FormatBuilder for SparqlBuilder {
...
+
+        if let Some(base) = base {
+            parser = parser.with_base_iri(base.to_string()).unwrap();
+        }
+
+        for (prefix_name, prefix_iri) in resolved_prefixes {
+            parser = parser
+                .with_prefix(prefix_name, prefix_iri.to_string())
+                .unwrap();
+        }
```
mmarx: It's maybe worthwhile to factor this into function that takes base and prefixes and returns a `SparqlParser`, since we're doing the same thing above

nemo/src/io/formats/sparql.rs
```diff
@@ -282,10 +339,20 @@ mod test {
...
+                "wikibase".to_string(),
+                IriRef::from_str("http://wikiba.se/ontology#").unwrap(),
+            ),
+            // PREFIX wdt: <http://www.wikidata.org/prop/direct/>
+            ("wdt".to_string(), IriRef::from_str("prop/direct/").unwrap()),
+        ]);
+
         let valid_query = AnyDataValue::new_plain_string(String::from("
-            PREFIX wikibase: <http://wikiba.se/ontology#>
+        PREFIX wikibase: <http://wikiba.se/ontology#>
```
mmarx, line 328: Not sure why we change the alignment here?

nemo/src/io/format_builder.rs
```diff
@@ -513,10 +530,20 @@ impl ImportExportBuilder {
...
+        let parameters = Parameters::<B>::validate(
+            predicate,
+            spec,
+            bindings,
+            filter_rules,
+            direction,
+            report,
+            base.clone(),
+            prefixes.clone(),
+        )?;
```
mmarx, line 539: This should probably be factored into one or possible several (that'd make it easier to reuse them elsewhere) structs. We can discuss the details in person tomorrow.

nemo/src/rule_model/translation/directive/import_export.rs
```diff
@@ -119,8 +123,19 @@ impl TranslationComponent for ImportDirective {
 
         let bindings = import_export_bindings(translation, import.guards())?;
 
+        let base = translation
+            .base
+            .clone()
+            .map(|x| Iri::from_str(&x.0).unwrap());
```
mmarx: Is the base already validated at this point? If yes, we can `expect` here, otherwise, this might need to produce an error.

nemo/src/rule_model/translation/directive/import_export.rs
```diff
@@ -119,8 +123,19 @@ impl TranslationComponent for ImportDirective {
...
+
+        let prefixes = translation
+            .prefix_mapping
+            .iter()
+            .map(|(key, value)| (key.clone(), IriRef::from_str(&value.0).unwrap()))
```
mmarx: Same question as for the base above.

mmarx (changes requested):
This is mostly fine, only a few stylistic issues remain

nemo/src/rule_model/translation/directive.rs
```diff
@@ -141,3 +145,73 @@ fn handle_prefix<'a>(
...
+                    .expect("relative prefix not resolvable on base");
+            } else {
+                panic!("attempted to resolve relative prefix without base")
+            }
+        }
+
+        self.prefixes.insert(prefix, abs_prefix);
+    }
+
+    ///
```
mmarx: This still needs a comment …

nemo/src/rule_model/translation/directive.rs
```diff
@@ -141,3 +145,73 @@ fn handle_prefix<'a>(
...
+    pub fn as_sparql_parser(&self) -> SparqlParser {
```
mmarx: As per the [API Naming Guidelines](https://rust-lang.github.io/api-guidelines/naming.html), `as_` indicates that a conversion is free – this isn't, so it should rather be called `to_sparql_parser`

nemo/src/rule_model/components/import_export.rs
```diff
@@ -387,9 +397,15 @@ impl ImportDirective {
             self.filter_rules(),
             Direction::Import,
             &mut report,
+            self.format_context().clone(),
         )
     }
 
+    /// Return the FormatContext
```
mmarx, line 403: we should link the `FormatContext` here: `[FormatContext]`

nemo/src/rule_model/components/import_export.rs
```diff
@@ -609,9 +627,15 @@ impl ExportDirective {
             self.filter_rules(),
             Direction::Export,
             &mut report,
+            self.format_context().clone(),
         )
     }
 
+    /// Return the FormatContext
```
mmarx, line 633: see above

nemo/src/rule_model/translation/directive.rs
```diff
@@ -141,3 +145,73 @@ fn handle_prefix<'a>(
         }
     }
 }
+
+/// Wrapper for IRI base and prefixes
+/// Automatically attempts to resolve all relative prefixes on the base
```
mmarx, line 175: This should be more generic towards what it actually holds: “Additional context for import/export formats (e.g., base and prefixes)” or something like this, since this might evolve to contain even more in the future. The part about resolving prefixes should probably rather be on the `add_prefix` method.

nemo/src/rule_model/translation/directive.rs
```diff
@@ -141,3 +145,73 @@ fn handle_prefix<'a>(
...
+impl FormatContext {
+    pub fn base(&self) -> &Option<Iri<String>> {
+        &self.base
+    }
+
+    pub fn prefixes(&self) -> &HashMap<String, Iri<String>> {
+        &self.prefixes
+    }
+
+    /// adds a base if None was set before
```
mmarx, line 193: We generally start comments with an uppercase letter (they should form a complete sentence) – this applies in some other places in the PR as well.

nemo/src/rule_model/translation/directive.rs
```diff
@@ -141,3 +145,73 @@ fn handle_prefix<'a>(
...
+    /// panics if a base is already set
```
mmarx, line 194: this should also mention that it panics if the given `base` is not a valid (absolute) IRI

nemo/src/rule_model/translation/directive.rs
```diff
@@ -141,3 +145,73 @@ fn handle_prefix<'a>(
...
+                abs_prefix = base
+                    .resolve(&rel_prefix.clone())
+                    .expect("relative prefix not resolvable on base");
+            } else {
+                panic!("attempted to resolve relative prefix without base")
+            }
+        }
+
+        self.prefixes.insert(prefix, abs_prefix);
+    }
```
mmarx: This can be just
```rust
        let abs_prefix = if rel_prefix.is_absolute() {
            Iri::parse(iri.clone()).expect("absolute IRI is not absolute")
        } else if let Some(base) = &self.base {
            base.resolve(&rel_prefix.clone())
                .expect("relative prefix not resolvable on base")
        } else {
            panic!("attempted to resolve relative prefix without base");
        };
```

mmarx (changes requested):
A few more things, again mostly stylistic in nature. Also a lot of unneeded imports that `cargo clippy --fix` should be able to get rid of …

mmarx (changes requested):
A few more things, again mostly stylistic in nature. Also a lot of unneeded imports that `cargo clippy --fix` should be able to get rid of …

nemo/src/rule_model/error/translation_error.rs
```diff
@@ -126,6 +126,17 @@ pub enum TranslationError {
     #[assoc(note = "parameter names must have the form `$name'")]
     #[assoc(code = 130)]
     ParamDeclarationNotGlobal,
+    /// Base is not a valid IRI
+    #[error(r#"base is not a valid IRI"#)]
+    #[assoc(note = "base must be a valid IRI (RFC 3987)")]
```
mmarx: at least in the note, we should mention that it should be an absolute IRI

nemo/src/rule_model/error/translation_error.rs
```diff
@@ -126,6 +126,17 @@ pub enum TranslationError {
...
+    #[assoc(code = 131)]
+    BaseInvalid,
+    /// Prefix is not a valid IRI
+    // should this be split into 2 (invalid ref/ not resolvable)?
+    #[error(r#"prefix is not a valid IRI"#)]
+    #[assoc(note = "prefix must be a valid IRI (RFC 3987)")]
```
mmarx: It's probably fine not to split this, as long as the note explains that it must resolve to an absolute IRI (i.e., potentially requires a base)

nemo/src/rule_model/translation/directive.rs
```diff
@@ -100,6 +99,12 @@ fn handle_base<'a>(translation: &mut ASTProgramTranslation, base: &ast::directiv
             .add_context_source(first_base.clone(), Info::FirstDefinition);
     }
 
+    // check if base is a valid iri
+    if Iri::parse(base.iri().content()).is_err() {
+        translation.report.add(base, TranslationError::BaseInvalid);
+        // TODO add parse error as context
```
mmarx, line 105: what's left to be done here?

nemo/src/rule_model/translation/directive.rs
```diff
@@ -131,6 +136,26 @@ fn handle_prefix<'a>(
...
+            return Ok::<(), Box<dyn std::error::Error>>(());
+        } else if let Some(base) = &translation.base {
+            Iri::parse(base.0.to_string())
+                .expect("base is not a valid IRI")
+                .resolve(&rel_prefix)?;
+            return Ok::<(), Box<dyn std::error::Error>>(());
+        }
+        Err(Box::new(TranslationError::PrefixInvalid))
+    })()
+    .is_err()
```
mmarx: It's probably better if you move this closure into its own declaration instead of invoking it directly. This will also let you specify the return type only once:
```rust
let parse_prefix: FnOnce() -> Result<(), Box<dyn std::error::Error> { … };
if (parse_prefix().is_err() { … }
```

nemo/src/rule_model/translation/directive.rs
```diff
@@ -155,10 +180,12 @@ pub struct FormatContext {
 }
 
 impl FormatContext {
+    /// Returns an optional base IRI
```
mmarx: probably rather something like “Returns the base IRI, if set”

nemo/src/rule_model/translation/directive.rs
```diff
@@ -180,24 +207,24 @@ impl FormatContext {
     /// - a relative prefix is given but no base is set
     pub fn add_prefix(&mut self, prefix: String, iri: String) {
         let rel_prefix = IriRef::parse(iri.clone()).expect("not a valid Iri");
-        let abs_prefix;
-        if rel_prefix.is_absolute() {
-            abs_prefix = Iri::parse(iri.clone()).expect("absolute IRI is not absolute");
+        let abs_prefix = if rel_prefix.is_absolute() {
+            Iri::parse(iri.clone()).expect("absolute IRI is not absolute")
```
mmarx: We're using the [“expect as precondition”](https://doc.rust-lang.org/std/error/index.html#common-message-styles) style, so this should read “IRI is absolute”

nemo/src/rule_model/translation/directive.rs
```diff
@@ -180,24 +207,24 @@ impl FormatContext {
...
         } else {
             if let Some(base) = &self.base {
-                abs_prefix = base
-                    .resolve(&rel_prefix.clone())
-                    .expect("relative prefix not resolvable on base");
+                base.resolve(&rel_prefix.clone())
+                    .expect("relative prefix not resolvable on base")
```
mmarx: see above, “relative prefix can be resolved using base”

nemo/src/rule_model/translation/directive.rs
```diff
@@ -180,24 +207,24 @@ impl FormatContext {
...
             } else {
-                panic!("attempted to resolve relative prefix without base")
+                panic!(
```
mmarx: this should be `unreachable!`

nemo/src/io/format_builder.rs
mmarx, line 520: Just add `#[allow(clippy::too_many_arguments)]` for now.

mmarx (changes requested):
A few more style nitpicks, but this is looking very good already!

nemo/src/rule_model/translation/directive.rs
Aeternant: The thought was to add the iri translation error to the report but I'm unsure if that is necessary. I have removed the TODO for now.

nemo/src/rule_model/translation/directive.rs
Aeternant: rust doesn't allow `impl Trait` for variable types, so return type specification isn't possible unfortunately.
(I moved the closure to a seperate variable though)

mmarx (approved):

PR #788, opened 2026-05-19
Trace all IDB facts

### LINKED ISSUE #783: Getting traces for all derived facts but without doing reasoning twice
Dear Nemo team,

Some time ago I asked a similar [question](https://github.com/knowsys/nemo/issues/668) and the answer was that I had to run `nmo` twice: 

1. Run `nmo` to determine  derived facts.
2. Run `nmo` again passing derived facts as facts to trace.

Is this still the only way to obtain traces for all derived facts? This, if I understand this correctly, means reasoning is done twice. If reasoning indeed is being done twice, is there any way to avoid this?

mmarx:
This adds an `--trace-all-idb-facts` CLI option (mutually exclusive with `--trace` and `--trace-input-file`) that will output traces for all derived facts after reasoning.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "crane": {
       "locked": {
-        "lastModified": 1776635034,
-        "narHash": "sha256-OEOJrT3ZfwbChzODfIH4GzlNTtOFuZFWPtW7jIeR8xU=",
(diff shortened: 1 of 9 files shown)


monsterkrampe (approved):
Looks good. Also, thanks for cleanup up the existing code!

PR #789, opened 2026-06-04
Bump flake to nixpkgs-26.05

mmarx:


flake.nix
```diff
@@ -2,7 +2,7 @@
   description = "nemo, a datalog-based rule engine for fast and scalable analytic data processing in memory";
 
   inputs = {
-    nixpkgs.url = "github:NixOS/nixpkgs/nixos-25.11";
+    nixpkgs.url = "https://channels.nixos.org/nixos-26.05/nixexprs.tar.xz";
```
monsterkrampe, line 5: Is there a particular reason for using this over `github:NixOS/nixpkgs/nixos-26.05`? (Genuine question since I might need to adjust this in my flakes too.) :D

monsterkrampe (approved):
Thanks for bumping us to the next version! I have one question but this is mainly out of curiosity :)

flake.nix
mmarx: It's served from the NixOS CDN instead of from GitHub, so it's quite a bit faster in practice, especially since it doesn't do a (shallow) git clone of the repository. It used to be the case that `flake.lock` didn't support locking the channels, but that's been in lix (and nix) for quite a few releases now.

mmarx: Cf. https://discourse.nixos.org/t/questions-about-inputs-nixpkgs/69095/25

monsterkrampe: Nice, thanks! :)

Issue #774, opened 2026-03-13
Transparently handle transformations in tracing


mmarx:
After #773 is implemented, we should trace rule origins back towards the untransformed program.

mkroetzsch:
Yes, but I don't see why this has to be strictly after #773. As it was conceived, the transformation origin system should also be usable when origin information is incomplete (as it might always be in this extensible transformation system). Of course, there should be some working cases to try it out on ... but it would actually be good to ensure that the back-transformation works with imperfect origin data as well.

PR #790, opened 2026-06-26
Allow Translation of Nev Rule Ids back to Rules and their Line Numbers

monsterkrampe:
This PR adds a function to `nemo-wasm` that allows to translate a rule id from Nev into the corresponding line number in the underlying Nemo program. For this to work, the PR also touches the handling of rule ids in the `nemo` crate.

@aannleax I'm not sure if you still wanted to clean things up here. For example, there is still a TODO in `rule_translation.rs`. If we cannot resolve this now then please create a technical debt issue for that. Once you are done, please mark the PR as ready :)


aannleax:
"If we cannot resolve this now then please create a technical debt issue for that."
I guess the issue exists already (https://github.com/knowsys/nemo/issues/774), though its a bit vague.

monsterkrampe:
Ah, ok. Maybe putting a link to the issue in the TODO comment then would be nice :)

nemo/Cargo.toml
```diff
@@ -18,6 +18,7 @@ js = ["getrandom/wasm_js"]
 nemo-physical = { path = "../nemo-physical", default-features = false }
 log = { workspace = true }
 nom = "7.1.1"
+line-index = "0.1.1"
```
mmarx: We already use that in `nemo-language-server`, so it's now time to move this out into the top-level `Cargo.toml` and inherit the version both here and there.

aannleax:
This PR also introduces RuleIdTranslation, which basically translates rule indices from the original (before transformations) to the normalized program. Hence, we assume now that no program transformation changes the structure of derivation trees (which is true now).

Also, it properly sets the origins for the global, incermental import, and sparql merge transformations (which is also covered by unit tests now).

And finally, some small changes to the API and how the python and wasm code builds programs and ExecutionEngines.

nemo-python/src/lib.rs
```diff
@@ -411,8 +412,11 @@ impl NemoEngine {
             .build()?;
 
         let execution_parameters = ExecutionParameters::default();
+
+        // Build the engine from the program's existing handle, so that parsing,
+        // to keep the same program pipeline
```
mmarx: This sentence looks broken

nemo/src/execution/tracing/rule_translation.rs
```diff
@@ -0,0 +1,250 @@
...
+        // an original rule, a transformation created a rule without an origin
+        // back-pointer, and we fail loudly rather than return a wrong rule id.
+        let norm_to_orig = normalized
+            .rules()
+            .iter()
+            .map(|rule| {
+                let origin_id = tracing_resolve_origin_id(handle, rule.id());
+                *orig_index_by_id.get(&origin_id).expect(
+                    "a normalized rule has no corresponding original rule; a \
+                     transformation created a rule without recording its origin",
```
mmarx: We are still using [“expect as a precondition” style](https://doc.rust-lang.org/std/error/index.html#common-message-styles), this is clearly an error message, which it should not be.

nemo/src/execution/tracing/rule_translation.rs
```diff
@@ -0,0 +1,250 @@
...
+            .map(|(index, rule)| (rule.id(), index))
+            .collect::<HashMap<_, _>>();
+
+        // For each normalized rule, resolve the original rule it should be
+        // displayed as and record that rule's index in the original program.
+        //
+        // This relies on the transformations being one-to-one on rules and
+        // recording their provenance: if a normalized rule cannot be resolved to
+        // an original rule, a transformation created a rule without an origin
+        // back-pointer, and we fail loudly rather than return a wrong rule id.
```
mmarx: This restates the function-level comment _and_ the precondition that should be in the expect() (see below). Beyond that, it provides no value, so why do we need a comment for that?

nemo/src/execution/tracing/rule_translation.rs
```diff
@@ -0,0 +1,250 @@
...
+                *orig_index_by_id.get(&origin_id).expect(
+                    "a normalized rule has no corresponding original rule; a \
+                     transformation created a rule without recording its origin",
+                )
+            })
+            .collect::<Vec<usize>>();
+
+        // Invert the mapping. As the transformations are one-to-one on rules, no
+        // two normalized rules map to the same original rule, so every insert is
+        // into a fresh slot (checked via the resulting length below).
```
mmarx: This restates the assumption _once again_. Please Dont.

It should also not be talking about slots, if anything, the correct term here is `element`.

nemo/src/execution/tracing/rule_translation.rs
```diff
@@ -0,0 +1,250 @@
...
+        let mut orig_to_norm = HashMap::with_capacity(norm_to_orig.len());
+        for (norm_index, &orig_index) in norm_to_orig.iter().enumerate() {
+            orig_to_norm.insert(orig_index, norm_index);
+        }
```
mmarx: Why not build both mapping simultaneously?

```rust
let mut norm_to_orig = HashMap::with_capacity(normalized.rules().count());
let mut orig_to_norm = HashMap::with_capacity(original.rules().count());

for rule in normalized.rules() {
    let origin_id = tracing_resolve_origin_id(handle, rule.id());
    let origin_index = orig_index_by_id.get(&origin_id).expect("each normalized rule has a corresponding original rule");
    let prev = norm_to_orig.insert(rule.id(), origin_index);
    debug_assert!(prev.is_none(), "each normalized rule maps to exactly one original rule");
    let prev = orig_to_norm.insert(origin_index, rule.id());
    debug_assert!(prev.is_none(), "each original rules maps to exactly one normalized rule");
}
```

nemo/src/execution/tracing/rule_translation.rs
```diff
@@ -0,0 +1,250 @@
...
+    /// rule was (transitively) derived from.
+    ///
+    /// # Panics
+    /// Panics if `normalized` is not a valid normalized rule index.
+    pub(crate) fn original_rule(&self, normalized: usize) -> Rule {
+        self.original
+            .rule_by_index(self.norm_to_orig[normalized])
+            .expect("normalized rule maps to an existing original rule")
+            .clone()
+    }
```
mmarx, line 90: This really shouldn't panic. We can have a panicking `original_rule_unchecked` version, if necessary.

nemo/src/execution/tracing/rule_translation.rs
```diff
@@ -0,0 +1,250 @@
...
+    pub(crate) fn to_original(&self, normalized: usize) -> usize {
+        self.norm_to_orig[normalized]
+    }
+
+    /// Translate an original (frontend) rule index into the normalized rule index used internally.
+    ///
+    /// # Panics
+    /// Panics if `original` has no corresponding normalized rule.
+    pub(crate) fn to_normalized(&self, original: usize) -> usize {
+        self.orig_to_norm
```
mmarx: `to_` signifies an [expensive ad-hoc conversion](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv). This is neither expensive, nor does it perform conversion, so it shouldn't be called like that. `original_index` and `normalized_index`, maybe?

nemo/src/execution/tracing/rule_translation.rs
```diff
@@ -0,0 +1,250 @@
...
+        assert_eq!(translation.original_rule(0).to_string(), original_string);
+        assert_ne!(translation.original_rule(0).to_string(), rewritten_string);
+    }
+
+    #[test]
+    #[cfg_attr(miri, ignore)]
+    fn translation_global_variable_rule() {
+        // The global variable `$t` is substituted by TransformationGlobal.
+        let translation = translation_for(
+            "@parameter $t = 5 .\ndata(5) .\nresult(?x) :- data(?x), ?x = $t .\n@output result .",
```
mmarx: Please write these as multi-line string literals.

nemo/src/execution/tracing/rule_translation.rs
```diff
@@ -0,0 +1,250 @@
...
+            r#"@import ra :- sparql {endpoint = "http://example.org", query = "SELECT ?a ?b WHERE {?a ?b ?z.}"} .
+            @import rb :- sparql {endpoint = "http://example.org", query = "SELECT ?a ?c WHERE {?a ?c ?z.}"} .
+            seed(1) .
+            result(?a, ?b, ?c) :- seed(?a), ra(?a, ?b), rb(?a, ?c) .
+            @output result ."#,
+        );
+
+        assert_eq!(translation.to_original(0), 0);
+        assert_eq!(translation.to_normalized(0), 0);
+    }
```
mmarx: The only thing these tests are checking* is that the normalized programs has at least one rule, which corresponds to the first rule of the original program. This is also already ensured by the test above (which also checks that the translated rule is actually different).

This provides no value, and these tests should therefore go.

*: except for the first one, which also does a useless check that the original rule contains some expected text with the unverified assumption that this is not also in the normalized rule.

nemo/src/execution/tracing/rule_translation.rs
```diff
@@ -0,0 +1,250 @@
...
+
+    /// Build a [RuleIdTranslation] for a program string
+    fn translation_for(program: &str) -> RuleIdTranslation {
+        let handle = crate::api::load_program_handle(program.to_string(), String::default())
+            .expect("program should parse and validate");
+        let normalized = NormalizedProgram::normalize_program(&handle);
+
+        RuleIdTranslation::new(&handle, &normalized)
+    }
```
mmarx: These can go once we get rid of the useless tests they're used in.

nemo/src/execution/execution_engine.rs
```diff
@@ -339,6 +349,20 @@ impl<Strategy: RuleSelectionStrategy> ExecutionEngine<Strategy> {
...
+    /// To obtain the original program, use [ProgramHandle::original_revision].
+    pub fn program_handle(&self) -> &ProgramHandle {
+        &self.program_handle
+    }
+
+    /// Return the [ProgramHandle] of the original program, i.e. the program as
+    /// it was created from the user input (the first revision of the pipeline).
+    pub fn original_program_handle(&self) -> ProgramHandle {
+        self.program_handle.original_revision()
+    }
```
mmarx: Either the comment on `program_handle` should refer to `original_program_handle`, or `original_program_handle` should just go away. Having both is confusing.

nemo/src/execution/execution_engine.rs
```diff
@@ -444,24 +468,46 @@ mod test {
...
-        api::load_program,
         execution::{DefaultExecutionEngine, execution_parameters::ExecutionParameters},
+        rule_file::RuleFile,
     };
 
+    #[tokio::test]
+    #[cfg_attr(miri, ignore)]
+    async fn rule_source_position_reports_line() {
+        // A fact on line 1 and a rule on line 2.
+        let source = "a(1).\nb(?x) :- a(?x).";
```
mmarx: Please write this using a multi-line string literal.

nemo/src/rule_model/pipeline/transformations/global.rs
```diff
@@ -88,6 +89,13 @@ impl<'a> ProgramTransformation for TransformationGlobal<'a> {
                 let mut new_statement = statement.clone();
                 substitution.apply(&mut new_statement);
 
+                // Record that this rule was derived from the original rule, so
+                // tracing can resolve it back and the rule-index translation
+                // between the original and the normalized program stays 1:1.
+                if let Statement::Rule(rule) = &mut new_statement {
+                    rule.set_origin(Origin::Global(statement.id()));
+                }
+
```
mmarx: We should track this for all statements, not merely for rules. If the statement is an import, it can still contain a parameter that gets replaced, and we want to keep track of that as well.

nemo/src/rule_model/pipeline/transformations/incremental.rs
```diff
@@ -309,13 +310,18 @@ impl ProgramTransformation for TransformationIncremental {
...
                         rule,
                         &incremental_predicates,
                         &derived_predicates,
                         &mut name_id,
                     );
 
+                    // Record that this rule was derived from the original rule, so
+                    // tracing can resolve it back and the rule-index translation
+                    // between the original and the normalized program stays 1:1.
+                    new_rule.set_origin(Origin::Incremental(rule.id()));
```
mmarx: Do we really need a 3-line comment on a single-line function call that merely restates _what the code does anyway_?

… we did not, after all :3

nemo/src/rule_model/programs/handle.rs
```diff
@@ -100,6 +113,32 @@ impl ProgramHandle {
     pub fn rule_by_id(&self, id: ProgramComponentId) -> Option<&Rule> {
         self.pipeline.rule_by_id(id)
     }
+
+    /// Return the [Rule] at the given index, counting rules only.
+    ///
+    /// This is the index used to refer to rules in tracing
+    /// (see [crate::execution::tracing::shared::RuleId]).
+    pub fn rule_by_index(&self, index: usize) -> Option<&Rule> {
```
mmarx: Instead of into the comment, the `RuleId` should go into the arguments list.

nemo/src/rule_model/origin.rs
```diff
@@ -38,7 +38,20 @@ pub enum Origin {
...
+    /// Rule that was created by substituting global variables in a rule
+    /// (see [crate::rule_model::pipeline::transformations::global::TransformationGlobal]).
+    Global(ProgramComponentId),
+
+    /// Rule that was created by inlining incremental imports into a rule
+    /// (see [crate::rule_model::pipeline::transformations::incremental::TransformationIncremental]).
+    Incremental(ProgramComponentId),
+
+    /// Rule that was created by merging SPARQL imports in a rule
+    /// (see [crate::rule_model::pipeline::transformations::merge_sparql::TransformationMergeSparql]).
```
mmarx: I'd prefer not to have all these references to particular places of the code where this is currently assumed to originate here. These will get stale once we rename transformations, and there's no guarantee that the named transformations will only ever be the only origin of these variants. The “find references” feature of any IDE does the same job as this comment, but better, without ever getting out of date or needing updates.

nemo/src/rule_model/error.rs
```diff
@@ -406,7 +406,10 @@ impl ValidationReport {
                 error.add_hint(Info::DefinedExternally);
                 None
             }
-            Origin::Normalization(id) => Self::id_to_range(program, *id, error),
+            Origin::Normalization(id)
+            | Origin::Global(id)
+            | Origin::Incremental(id)
+            | Origin::MergeSparql(id) => Self::id_to_range(program, *id, error),
```
mmarx, line 412: Perhaps it's better to have a generic `Origin::Rewriting(id, kind)`, where `kind` is an enum that details the concrete rewriting used? That'd certainly alleviate quite a bit of friction for adding more transformations.

nemo/src/rule_model/pipeline.rs
```diff
@@ -41,6 +47,34 @@ impl ProgramPipeline {
...
+    /// Return a [ProgramHandle] to the original revision of the [ProgramPipeline],
+    /// i.e. the first revision. This holds the program as it was created from the
+    /// user input, before any transformation.
+    ///
+    /// Every revision derived from this pipeline shares the same first revision,
+    /// so this is the single reference point for resolving components back to the
+    /// original program (see the tracing modules).
+    pub fn original_revision(self: &Rc<Self>) -> ProgramHandle {
+        debug_assert!(
+            self.revisions.len() > 0,
```
mmarx: `!self.revisions.is_empty()`? I guess [clippy will flag it anyways](https://rust-lang.github.io/rust-clippy/master/index.html?search=empty#len_zero) …

nemo/src/parser/span.rs
```diff
@@ -358,3 +401,48 @@ impl nom::Compare<&str> for Span<'_> {
...
+
+        // The single character on the fourth line.
+        let f = positions.position(8);
+        assert_eq!((f.line, f.column), (4, 1));
+        assert_eq!(f.offset, 8);
+    }
+
+    #[test]
+    fn source_positions_range() {
+        let source = "a(1).\nb(?x) :- a(?x).";
```
mmarx: Please use a multi-line string literal here.

nemo/src/parser/span.rs
```diff
@@ -358,3 +401,48 @@ impl nom::Compare<&str> for Span<'_> {
...
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::SourcePositions;
+
+    #[test]
+    fn source_positions_line_and_column() {
+        let source = "abc\nde\n\nf";
```
mmarx, line 411: Please use a multi-line string literal here.

nemo/src/rule_model/pipeline.rs
```diff
@@ -41,6 +47,34 @@ impl ProgramPipeline {
...
+        );
+
+        ProgramHandle::new(self.clone(), 0)
+    }
+
+    /// Record the source text the program was parsed from.
+    ///
+    /// Has no effect if a source has already been set.
+    pub fn set_source(&self, source: &str) {
+        let _ = self.source.set(SourcePositions::new(source));
```
mmarx: Since setting this twice is something we would like to discover, perhaps
```rust
let set = self.source.set(SourcePositions::new(source));
debug_assert!(set.is_ok(), "source position is not already set");
```

mmarx (changes requested):

nemo/src/execution/tracing/rule_translation.rs
aannleax: I can just call it unchecked. There is no reason to use the checked version because there is no way to recover from such an error

nemo/src/execution/tracing/rule_translation.rs
aannleax: I mean the important thing is it checks is that the origin resolution works. Note that before this PR such a test would have failed.

nemo/src/execution/execution_engine.rs
aannleax: hm maybe fn program_handle should be named current_program_handle to emphasize this?

nemo/src/execution/tracing/rule_translation.rs
mmarx: Yes, but that is also already done by the first test (which I have not marked for removal). Beyond that, these four tests add nothing, except for longer CI runs and the potential for breaking due to unrelated changes.

mmarx: What _would_ be a useful test is one where we have more than a single rule, e.g..

nemo/src/execution/tracing/rule_translation.rs
mmarx: For now, maybe not, but this could be called interactively, e.g. (but we can add an `original_rule` then, once we can use it, yes).

nemo/src/execution/execution_engine.rs
mmarx: That is not at all what I mean. What I mean is either:
```rust
 /// To obtain the original program, use [ProgramHandle::original_revision].
    pub fn program_handle(&self) -> &ProgramHandle {
        &self.program_handle
    }

// no original_program_handle() here
```

or
```rust
    /// To obtain the original program, use [Self::original_program_handle].
    pub fn program_handle(&self) -> &ProgramHandle {
        &self.program_handle
    }

    /// Return the [ProgramHandle] of the original program, i.e. the program as
    /// it was created from the user input (the first revision of the pipeline).
    pub fn original_program_handle(&self) -> ProgramHandle {
        self.program_handle.original_revision()
    }
```

nemo/src/rule_model/error.rs
aannleax: Hm that is an interesting suggestion. My original intention was to have one enum variant correspond to a transformation. The point would be that you are forced to consider how to properly reverse in each scenario. 

For that reason I also think RuleRewriting might be to generic. Depending on how the rule is rewritten (e.g. does it preserve variables), reverting it may require a different approach. (In this case they behave the same though)

nemo/src/execution/tracing/rule_translation.rs
mmarx: What _might_ also be a useful test is that the transformations indeed create rules with the correct origins set, but that is absolutely not a unit test on `RuleIdTranslation`; rather, these should (i) be tests alongside the actual translations, and (ii) check that they actually set the correct origin (which is what we actually care about, not just that they have _some_ origin mapping them back to the original rule).

nemo/src/rule_model/error.rs
mmarx: Ah yes, that is a good point. Let's keep it like this, then.

nemo/src/rule_model/programs/handle.rs
aannleax: Probably the comment should just go; this function could potentially be used in other situations as well

nemo/src/execution/tracing/rule_translation.rs
aannleax: Yes, agreed

nemo/src/execution/tracing/rule_translation.rs
aannleax: I see your point somewhat. 

The point of these tests was to ensure that the current transformations adhere to the tree transformation paradigm set my this RuleIdTranslation (i.e. injective rule-to-rule mappings). This is also the reason why they exist here. 

But I suppose I can't force people to write such tests for new transformations, so the guarentee this gives is somewhat limited

nemo/src/execution/execution_engine.rs
aannleax: ah yes of course

nemo/src/execution/tracing/rule_translation.rs
mmarx: That's also what we have the `debug_assert!` for, isn't it?

These tests are fundamentally integration tests. Several parts of this we can test using unit tests, and some we cannot: 
* that the mapping in general works (that's already ensured by the test above)
* that each of the transformations sets the correct `origin` (this is, notably, not covered by these tests – they would pass if `TransformationGlobal` would set `Origin::Incremental`, e.g.; and that's also fundamentally a test for the individual transformations, because it _tests the behaviour of the transformation, not the behaviour of `RuleIdTransformation`_
* there should be no transformation that creates a non-injective rule mapping (this would trigger the `debug_assert`, but it's fundamentally impossible to write a test case for this, since new transformations can always be added which simply don't conform to this)

We might consider upgrading the `debug_assert!` to a proper `assert!` and just panic, since there's no real guarantee that non-conforming transformations get triggered in debug builds (there might be transformations targeting only very large programs, e.g., that wouldn't be run under debug builds). The overhead for this is pretty negligible, but at least we don't silently report wrong results in those cases.

aannleax: "That's also what we have the debug_assert! for, isn't it?"
Ok, but this would only trigger if you ran nemo on such a rule set in debug mode...

I would like to catch this earlier. So I want to check at test time whether this debug assert fires

mmarx: I guess what you rather want to check is that the default transformations _do not trigger_ the `debug_assert!`, and that's covered by the two other kinds of tests I described?

nemo/src/parser/span.rs
aannleax: for this test I feel like the single line approach is more readable. I adjusted the other ones.

nemo/src/execution/tracing/rule_translation.rs
aannleax: ok this makes more sense to me

aannleax: normalize, global, incremental and merge sparql now have unit tests verifying that the origin has been set.

And the default transformation have received a unit test which contains a program that should trigger all problematic transformations and check whether RuleIdTranslation not panics and translates them back correctly.

nemo/src/rule_model/pipeline/transformations/global.rs
```diff
@@ -97,3 +99,37 @@ impl<'a> ProgramTransformation for TransformationGlobal<'a> {
...
+                .into_object();
+
+        let original = handle.rules().next().unwrap().id();
+
+        let transformed = handle
+            .transform(TransformationGlobal::new(&HashMap::new()))
+            .unwrap();
+        let rule = transformed.rules().next().unwrap();
+
+        assert!(matches!(rule.origin(), Origin::Global(id) if id == original));
```
mmarx: Better use `assert_matches!` instead of `assert!(matches!(…))`, that'll give much better error messages in case of failure:
```rust
use std::assert_matches;

assert_matches!(rule.origin(), Origin::Global(id) if id == original)
```

nemo/src/rule_model/pipeline/transformations/incremental.rs
```diff
@@ -332,3 +335,36 @@ impl ProgramTransformation for TransformationIncremental {
...
+            ProgramHandle::from_file(&RuleFile::new(program.to_string(), String::default()))
+                .expect("program parses")
+                .into_object();
+
+        let original = handle.rules().next().unwrap().id();
+
+        let transformed = handle.transform(TransformationIncremental::new()).unwrap();
+        let rule = transformed.rules().next().unwrap();
+
+        assert!(matches!(rule.origin(), Origin::Incremental(id) if id == original));
```
mmarx: (see above)

nemo/src/rule_model/pipeline/transformations/merge_sparql.rs
```diff
@@ -268,3 +272,39 @@ fn update_import(
...
+                .expect("program parses")
+                .into_object();
+
+        let inlined = handle.transform(TransformationIncremental::new()).unwrap();
+        let predecessor = inlined.rules().next().unwrap().id();
+
+        let transformed = inlined.transform(TransformationMergeSparql).unwrap();
+        let rule = transformed.rules().next().unwrap();
+
+        assert!(matches!(rule.origin(), Origin::MergeSparql(id) if id == predecessor));
```
mmarx: (see above)

nemo/src/rule_model/pipeline/transformations/normalize.rs
```diff
@@ -148,3 +148,35 @@ impl ProgramTransformation for TransformationNormalize {
...
+                .into_object();
+
+        let original = handle.rules().next().unwrap().id();
+
+        let transformed = handle
+            .transform(TransformationNormalize::default())
+            .unwrap();
+        let rule = transformed.rules().next().unwrap();
+
+        assert!(matches!(rule.origin(), Origin::Normalization(id) if id == original));
```
mmarx: (see above)

mmarx (changes requested):
Yeah, much better. I'd just prefer `assert_matches!(…)` over `assert!(matches!(…))`, since that gives much better error messages.

mmarx (approved):

PR #797, opened 2026-08-04
Improve Performance of Variable Order Computation

aannleax:
After the dependency graph computation, the variable order computation had the highest impact on large rule sets. This PR improves the performance by reducing memory allocations and rewriting the algorithm.

### DIFF
--- a/nemo/src/execution/planning/analysis/variable_order.rs
+++ b/nemo/src/execution/planning/analysis/variable_order.rs
@@ -167,8 +167,6 @@ impl std::fmt::Debug for VariableOrder {
     }
 }
 
-// The variants not listed in `ITERATION_ORDERS` are unused, but kept so that computing
-// several variable orders per rule can be reactivated by extending that list.
 #[allow(dead_code)]
 #[derive(Debug, Clone, Copy)]
 enum IterationOrder {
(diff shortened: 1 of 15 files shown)


nemo/src/execution/planning/analysis/variable_order.rs
```diff
@@ -170,13 +167,20 @@ impl std::fmt::Debug for VariableOrder {
     }
 }
 
-#[derive(Debug)]
+#[allow(dead_code)]
```
monsterkrampe, line 170: Maybe we should remove the variants that we do not need anymore instead of marking this as `dead_code`?
At least if you came to the conclusion that `Backward` iteration is preferred anyway, then I'm ok with removing the others.

monsterkrampe (approved):
Looks good. I did not verify how much faster this really is but I'm assuming that you tested that extensively :)

nemo/src/execution/planning/analysis/variable_order.rs
aannleax: Its the order which coincides with the one which we were previosly implcitly using (at least on the relevant benchmarks). I would just keep the old ones around in case we want to revisit this later.

monsterkrampe: Ah I see... Then let's keep it as is.

PR #801, opened 2026-08-09
Fix mixed numeric sum aggregation

### LINKED ISSUE #800: Mixed integer and float #sum drops float contribution
## Description

When `#sum` aggregates a group containing both integer and 32-bit float values, but no double value, Nemo silently returns only the integer subtotal. The float contribution is discarded and execution still succeeds.

## Reproduction

```prolog
input(1).
input(2.5f).

result(#sum(?value)) :- input(?value).
```

Run:

```console
cargo run -q -p nemo-cli -- --report none --print-facts idb mixed_sum.rls
```

Actual output:

```text
result(1).
```

Expected output, following the existing mixed-numeric promotion behavior:

```text
result("3.5"^^<http://www.w3.org/2001/XMLSchema#double>).
```

## Cause

`SumAggregateGroupProcessor::finish` checks `current_sum_f64` twice when deciding whether to combine numeric subtotals. For an integer/float group without a double, the combination branch is skipped and the fallback returns the integer subtotal before considering the float subtotal.

Reproduced on `main` at `daff6df7ce5968f5a4e0ec9079172ea415a8fd80`.

mattfaltyn:
## Summary
- combine integer and float subtotals as a double in `#sum`
- add a focused regression test for mixed integer/float input

## Testing
- `cargo test`
- `cargo fmt --all -- --check`
- `cargo clippy -p nemo-physical -p nemo --all-targets -- --deny warnings --allow clippy::nonstandard_macro_braces --allow clippy::double_must_use`

The two allowed clippy lints are existing diagnostics in unrelated files; the changed target is warning-free.

Fixes #800.


aannleax:
Looks good, thank you :)

aannleax:
I will fix the unrelated clippy error in another PR

aannleax (approved):

PR #804, opened 2026-08-12
Reject out-of-range numeric casts

### LINKED ISSUE #803: Out-of-range numeric casts panic or clamp
## Description

Numeric casts do not consistently reject values that cannot be represented by the target type:

- `FLOAT(1.0E100)` panics because converting the double to `f32` produces infinity and the implementation calls `expect`.
- `INT(1.0E20)` and `INT(-1.0E20)` silently clamp to `i64::MAX` and `i64::MIN` through Rust's saturating float-to-integer `as` conversion.

This contradicts the cast-function contract in `nemo-physical/src/function/definitions/casting.rs`, which says values outside the supported range or values that cannot be converted should return `None`.

This is distinct from #582/#588: those addressed arithmetic operations that produce non-finite values, while this issue concerns narrowing conversions in the `FLOAT` and `INT` builtins.

### `FLOAT` panic

```rls
input(1.0E100).
result(?Input, ?Float) :- input(?Input), ?Float = FLOAT(?Input).
```

```console
$ nmo cast-float.rls --print-facts idb --report none
thread 'main' panicked at nemo-physical/src/function/definitions/casting.rs:
resulting float must be finite: NonFiniteFloat
```

The command exits with status 101.

### `INT` clamping

```rls
input(1.0E20).
input(-1.0E20).
input(42.0E0).
result(?Input, ?Integer) :- input(?Input), ?Integer = INT(?Input).
```

Actual output:

```text
result("-100000000000000000000"^^<http://www.w3.org/2001/XMLSchema#double>, -9223372036854775808).
result("42"^^<http://www.w3.org/2001/XMLSchema#double>, 42).
result("100000000000000000000"^^<http://www.w3.org/2001/XMLSchema#double>, 9223372036854775807).
```

Both reproductions are deterministic on current `main` (`93f219c31f227f04fbde3a1ec272343a69c4c967`).

## Expected behavior

An out-of-range cast should evaluate to `None`, so no derived fact is produced and the process does not panic. Representable integral values such as `INT(42.0E0)` should continue to succeed.

## Root cause

`CastingIntoInteger64` checks whether a float has a fractional part but not whether it is in the `i64` range before using `as i64`. `CastingIntoFloat` narrows a double with `as f32` and then calls `expect` when constructing a finite Nemo float.

## Suggested validation

Add focused boundary tests covering:

- out-of-range positive and negative double-to-integer casts returning `None`;
- an out-of-range double-to-float cast returning `None` without panicking;
- nearby representable values continuing to convert successfully.

mattfaltyn:
## Summary

- reject out-of-range float and double inputs to `INT` instead of saturating them
- return no value when narrowing a double to `FLOAT` produces a non-finite value
- cover positive and negative overflow alongside representable control and boundary cases

## Why

Numeric casts are partial operations: inputs that the target datatype cannot represent should produce no value. Previously, Rust's `as i64` conversion silently clamped large floating-point values to `i64::MIN` or `i64::MAX`, while double-to-float overflow reached an `expect` and panicked the process.

The implementation now uses the existing checked numeric conversion for integer targets and propagates the float constructor's error as `None`.

Fixes #803.

## Testing

- `cargo test --workspace`
- `cargo test --workspace function::evaluation::test::evaluate_casting -- --exact`
- `cargo fmt --all -- --check`
- `RUSTFLAGS=-Dwarnings cargo clippy --all-targets --workspace --verbose`
- `RUSTFLAGS=-Dwarnings RUSTDOCFLAGS=-Dwarnings cargo doc --workspace`
- `pycodestyle nemo-python`
- built `nemo-cli` and reran both issue reproductions; invalid casts emit no fact, while a valid control still emits `42`

## Local environment limitations

- the full Miri suite reaches an existing Apple ARM limitation in `bytecount`: Miri does not support the invoked AArch64 NEON intrinsic
- a focused `nemo-physical` Miri run is blocked before this test by an existing `#[tokio::test]` expansion error on the local Apple/nightly toolchain
- Python wheel tests require `maturin`, which is not installed locally
- browser WASM tests require `wasm-pack` and configured Firefox/Chrome executables, which are not installed locally

The repository's Linux CI covers these platform-dependent jobs.


aannleax (approved):
Thank you looks good :)

PR #807, opened 2026-08-18
Reuse Skolem terms for repeated existential variables

xR0xEr:
Skolemization previously created a new function for every occurrence of an existential variable. This incorrectly allowed equal occurrences to denote different values.

Reuse one Skolem term per existential variable, retain transformed rules, and add a regression test covering repeated variables across head atoms.


monsterkrampe (approved):
Thanks, LGTM :)

PR #811, opened 2026-08-24
Experimental Jupyter kernel for Nemo language

### LINKED ISSUE #810: Add Jupyter kernel for Nemo
I like to do development in Jupyter notebooks.  I think that could help introduce folks to using Nemo.

jimwhite:
This fixes https://github.com/knowsys/nemo/issues/810 by adding nemo-jupyter, an experimental Jupyter kernel for Nemo language.

This is a quick vibe (prime-agent using DeepSeek V4 Flash) and I've only run the examples/hello-nemo.ipynb notebook but it works to that extent.  Also includes Export as Executable Script (.rls) support.

It runs in my .devcontainer setup, which is also included (and probably should be its own issue & PR).  

I realize this may not be suitable as-is for merging but I've made the PR for feedback (hoping it is close to something you'd like to include).


mmarx:
From a (very cursory) glance, I don't see why any of this should even live in the main repo. The Jupyter kernel just builds on the public Python API, so it can go into a downstream `nemo-jupyter` repo.

The submodules for `nemo-web` and `nemo-vscode-extension` don't belong here; the Dockerfile (i) doesn't even package Nemo, but rather `nemo-web` and (ii) looks like it can be replaced by a few lines of Nix code, since we've already got everything packaged; the devcontainer setup (i) is really just a devcontainer for the Jupyter kernel, not for Nemo and (ii) uses a horribly broken hack (that'll break when the underlying Python version changes, e.g.) to badly reinvent half of `mkPythonEditablePackage`.

The actual implementation of the Jupyter kernel also seems to have some issues, at the very least, the media type is invalid, the referenced pygments lexer doesn't exist, and the selection of completions is a somewhat random subset of keywords and builtins. 

Architecture-wise, it's probably better to build this atop `nemo-lanugage-server` (which could export a Python module, e.g.) instead of reinventing parsing and completion again.

jimwhite:
> From a (very cursory) glance, I don't see why any of this should even live in the main repo. The Jupyter kernel just builds on the public Python API, so it can go into a downstream `nemo-jupyter` repo.

Yeah, I was thinking about the same thing but started with this because those other parts, notably the nemo-python which this depends on, are here.

